PR #13, opened 2024-04-22
Feat/rustcbind

vilukissa68:
Adds example for building C source with bsp using cmake.


examples/c-bsp-test/linker.lds
```diff
@@ -0,0 +1,299 @@
+/* Copyright (c) 2020 SiFive Inc. */
+/* SPDX-License-Identifier: Apache-2.0 */
```
hegza, line 2: Apache-2.0: retain the License notice but also clearly state below what was changed from the original.

examples/c-bsp-test/RISC-V-Toolchain.cmake
```diff
@@ -0,0 +1,7 @@
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_SYSTEM_PROCESSOR riscv64)
+
+set(CMAKE_C_COMPILER /opt/homebrew/Cellar/riscv-gnu-toolchain/main/bin/riscv64-unknown-elf-gcc)
+set(CMAKE_ASM_COMPILER /opt/homebrew/Cellar/riscv-gnu-toolchain/main/bin/riscv64-unknown-elf-as)
+set(CMAKE_LINLKER /opt/homebrew/Cellar/riscv-gnu-toolchain/main/bin/riscv64-unknown-elf-ld)
```
hegza: Can we read these from path using conventional environment variables, e.g. "CC"? Alternatively, CMake also allows making variables "user-defined". The user should be prompted for these upon first running cmake.

hegza (changes requested):
Looks mighty fine. I think the license related thing is the only thing required for merge.

examples/c-bsp-test/linker.lds
vilukissa68: Does this look okay [3e9e41f](https://github.com/soc-hub-fi/headsail-vp/pull/13/commits/3e9e41f96ccce11bb8cb1f36e04a973b38eee66f)?

examples/c-bsp-test/RISC-V-Toolchain.cmake
vilukissa68: CC is now used. Toolchain file deleted.

vilukissa68:
For some reason the initial PR had rust-rt still enabled in the ffi. [81488b6](https://github.com/soc-hub-fi/headsail-vp/pull/13/commits/81488b6922677dbcc3656632c459576f0989b7d7) removes rust-rt from ffi by using the panic-uart features, which was previously deemed correct.

examples/c-bsp-test/linker.lds
hegza: Seems good enough

hegza (approved):
If you've tested it runs on sim, I think it's good to go.

Ideally we run this in CI as well but I can do that later.

PR #15, opened 2024-04-29
Feat/allocator

vilukissa68:
Adds allocator feature to HPC. This requires adding SDRAM to the HPC memory model.


examples/headsail-bsp/Cargo.toml
```diff
@@ -16,6 +16,7 @@ sysctrl-rt = [
 panic-uart = []
 hpc = []
 sysctrl = []
+allocator = ["dep:good_memory_allocator", "hpc"]
```
hegza: This feature is typically called `alloc` by convention.

examples/headsail-bsp/mem_hpc.x
hegza, line 1: I recommend splitting this link script to another file. Make one script that puts everything on SDRAM and another script that puts everything on boot RAM. This disambiguates "application level" programs from "bootloader" programs and helps with maintaining binary size limits when designing the boot stack later.

You'll need to update the reference in .cargo/config.toml and/or CMakeLists.txt where appropriate.

hegza (changes requested):

examples/headsail-bsp/Cargo.toml
vilukissa68: All references to allocator changed to alloc

examples/headsail-bsp/mem_hpc.x
vilukissa68: Added sdram feature to bsp that control the memory target during build. Modified build.rs file and cargo/config.toml in such a way that previous examples still work, but when using -Fsdram build targets sdram instead of bootram. Setting RUSTFLAGS manually when invoking build still works.

hegza:
Update with rebase to fix the FFI in CI. Looks about right otherwise. The link script configuration seems unconventional but that might be a personal bias of mine so we'll allow it :)

hegza (approved):

vilukissa68:
https://github.com/soc-hub-fi/headsail-vp/pull/18 should be merged before merging this PR so that the CI passes.

PR #20, opened 2024-05-06
Adds PULP timer_unit driver for timer0

andstepan:
This is the driver implements the interfacing for the timer peripheral found in the headsail-vp. The peripheral found in the VP is not the PULP APB Timer, as specified in the Headsail documentation, but is the PULP Timer Unit ([https://github.com/pulp-platform/timer_unit](https://github.com/pulp-platform/timer_unit)). This means, that **this driver is valid for the VP ONLY**.

Summary of changes:
examples/headsail-bp/Cargo.toml: Added timer_unit module dependencies and example entry
examples/headsail-bp/src/lib.rs: Added timer_unit module
examples/headsail-bp/src/mmap.rs: Added timer0 address constant
examples/headsail-bp/src/timer_unit.rs: Driver implementation (New file)
examples/hpc/hello-hpc/examples/timer0.rs: Example implementation (New file)


vilukissa68:
For some reason @andstepan doesn't have rights to run the CI, so I approved this run.

vilukissa68:
Tested example with:
```sh
cargo run --example timer0 --target riscv64imac-unknown-none-elf
```
Works well!

examples/headsail-bsp/Cargo.toml
```diff
@@ -49,3 +50,8 @@ required-features = ["rt"]
 name = "alloc"
 path = "examples/alloc.rs"
 required-features = ["alloc", "rt", "sdram"]
+
+[[example]]
+name = "timer0"
+path = "examples/timer0.rs"
+required-features = ["rt"]
```
hegza, line 58: This entry implies that there should be an example under project `headsail-bsp` at location `examples/headsail-bsp/examples/timer0.rs". In reality, the example is located under project `hello-hpc` at `examples/hpc/hello-hpc/examples/timer0.rs`.

Pick one:

- Move the exercise from hello-hpc/examples to headsail-bsp/examples (**preferred**)
- Move the [[example]] entry from headsail-bsp/Cargo.toml to hello-hpc/Cargo.toml

examples/headsail-bsp/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 pub mod sprintln;
 pub mod uart;
+pub mod timer_unit;
```
hegza, line 6: Since there may be several timers on the platform, an argument could be made that being very specific with the name might be a good idea, e.g. `pulp_timer_unit`. But after reconsidering, I do like `timer_unit` here more. I think its the perfect compromise of specificity and briefness.

examples/headsail-bsp/src/mmap.rs
```diff
@@ -1 +1,2 @@
 pub(crate) const UART0_ADDR: usize = 0xFFF00000;
+pub(crate) const TIMER0_ADDR: usize = 0x5_0000;
```
hegza, line 2: Thanks for adding the underscore here. I think it's generally a good idea to include those in the addresses.

examples/headsail-bsp/src/timer_unit.rs
```diff
@@ -0,0 +1,49 @@
+/**
+ * Date: 6/5/2024
+ * Author: Andreas Stergiopoulos (andreas.stergiopoulos@tuni.fi)
+ *
+ * This is the driver for the PULP timer unit. This driver is to
+ * be used in the VIRTUAL PROTOTYPE ONLY.
+ *
+ * Documentation: https://github.com/pulp-platform/timer_unit/tree/master
+ */
```
hegza, line 9: Excellent header comment. This makes it very clear that this is VP only.

I think the date is a bit superfluous, as git already takes care of that for us, and there will be changes to this driver later which will muddy the waters with that anyway.

examples/headsail-bsp/src/timer_unit.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::{mmap::TIMER0_ADDR, read_u32, write_u32};
+
+extern crate bit_field;
```
hegza: This `extern crate` is 2015 syntax, and is no longer necessary in the 2023 edition. Dependency crates are imported implicitly on first use. You may leave this out and the program will be  equivalent. Less lines is better.

examples/hpc/hello-hpc/examples/timer0.rs
hegza, line 1: This program uses the timer0_get_count which is only available in debug builds, but does not signal this in any way. I would prefer all programs can be compiled in `release` and `debug`, allowing for `cargo build --examples` to work in all  invocations without edge cases. Therefore, I recommend that you conditionally compile out the lines that use the debug-only features.

You can test everything is ok by running `cargo build --example timer0` and `cargo build --release --example timer0` and seeing that both complete without errors.

hegza: Alternatively, making timer0_get_count available in release builds looks like a valid option here

examples/headsail-bsp/src/timer_unit.rs
```diff
@@ -0,0 +1,49 @@
...
+use bit_field::BitField;
+
+const TIMER0_CTRL_REG_OFFSET: usize = 0x0;
+const TIMER0_COUNTER_REG_OFFSET: usize = 0x8;
+const TIMER0_ENABLE_BIT: u32 = 0b0;
```
hegza: The `TIMER0_ENABLE_BIT` constant is used as an indexer in all contexts, therefore its type should probably be expressed as `usize`. This will also allow you to remove the redundant type casts below.

examples/headsail-bsp/src/timer_unit.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[inline]
+pub fn timer0_enable() {
+    // Read register
+    let mut reg = read_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET);
+    // Make enable bit 1
+    reg.set_bit(TIMER0_ENABLE_BIT as usize, true);
+    // Write register back
+    write_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET, reg as u32);
```
hegza: `reg as u32` this looks like a redundant type conversion. We have already read a u32 from the register and we're putting a u32 back in. A type conversion should not be necessary.

hegza (changes requested):
Looks quite alright with minor changes. We might want to upgrade it to a relocatable HAL driver later, but that's a tomorrow's problem we don't need to be concerned with now.

Congratulations on your first Rust driver :smiley:

examples/headsail-bsp/src/lib.rs
hegza, line 6: One more thing! We might want to conditionally compile this line with `[cfg(feature = "vp")]` to make sure it's not included when targeting the ASIC. Then add the line `vp = []` under `[features]` in Cargo.toml, to declare the feature.

hegza: Ah, it's in the other PR. I would've put it in this one already but that's all fine. Consider this comment chain void.

examples/headsail-bsp/Cargo.toml
andstepan: Done, moved timer0.rs to headsail-bsp/examples.

examples/headsail-bsp/src/timer_unit.rs
andstepan: Done, removed line 12.

examples/hpc/hello-hpc/examples/timer0.rs
andstepan: Yes, I agree with your second comment, I think timer0_get_count() should probably be for release too as it can be used to profile execution times etc. However, the timer0_get_ctrl_reg() should probably remain debug only as I see no reason why would someone want to read the timer's control register during runtime. The user of the driver will probably just configure the timer and use it, never reading back what is in the control register.
I removed the debug_assertions from the timer0_get_count() function.

examples/headsail-bsp/src/timer_unit.rs
andstepan: My initial idea was that this should be u3 instead of usize since we are referring to registers and not pointers and those registers are always 32 bits wide. I have corrected the code, but it is still not very clear to me why usize should be used here instead of u32.

examples/headsail-bsp/src/timer_unit.rs
andstepan: Done, removed type conversion.

examples/headsail-bsp/src/timer_unit.rs
hegza: One superfluous cast still remains on line 39 but we can fix that later.

hegza: I think what is meant here with this constant is "which bit is our target bit, of the 32 options in the u32", with a value range of `0..32` (so technically u8 suffices if we wanted to optimize for space; though we do not).

We are using the value to do a bit-shift (even if that bit shift is encapsulated by the `bit_field` library) and the bit-shift operation takes a usize as a parameter. This is reflected in the API of the `bit_field` library which also takes a `usize` as the parameter in the `set_bit` function.

It's indeed typical that when a number is used as an indexer in a table such as an array of bits, it uses the native platform pointer width a.k.a. `usize`.

hegza (approved):
Looks good to me now. I'll let you press the squash & merge button in case there's something you'd like to change prior.

PR #22, opened 2024-05-06
Adds output address setting to DLA.py

vilukissa68:
Enables "hidden register" 5c which is used for setting the output address for the DLA. 
Includes accumulated minor fixes and changes.

### DIFF
--- a/vp/devel/python_peripherals/DLA.py
+++ b/vp/devel/python_peripherals/DLA.py
@@ -1,5 +1,3 @@
-import os
-import re
 import math
 import random
 
@@ -133,6 +131,10 @@ MAC_SAT_MAX_OFFSET = 0
 MAC_SAT_MIN = 0x58
 MAC_SAT_MIN_OFFSET = 0
 
+# PP_AXI_WRITE
+PP_AXI_WRITE = 0x5C
+PP_AXI_WRITE_ADDRESS_OFFSET = 0
+
 # PP_AXI_READ
 PP_AXI_READ = 0x60
 PP_AXI_READ_ADDRESS_OFFSET = 0
@@ -181,7 +183,6 @@ def flatten_tensor(data):
 
 def flat_to_CWH(data, channels, width, height):
     """Takes in 1d array of length C*W*H and reshapes it to tensor of format CWH"""
-    print(len(data), channels, width, height)
     assert channels * width * height == len(data)
     output = [[[0 for _ in range(height)] for _ in range(width)] for _ in range(channels)]
     i = 0
@@ -213,6 +214,9 @@ def separate_channels(data):
         channel_matrices.append(channel_matrix)
     return channel_matrices
 
+def bit_not(n, numbits=32):
+    return (1 << numbits) - 1 - n
+
 def print_matrix(A, name=""):
     """Print matrix"""
     print(name)
@@ -221,6 +225,9 @@ def print_matrix(A, name=""):
         for y in range(len(A[0])):
             row = row + str(A[x][y]) + " "
         print(row)
+
+def memory_bank_to_offset(bank):
+    return bank * MEMORY_BANK_SIZE
 
 def execute_for_all_elements(f, tensor, *args):
     """Execute function that takes a single matrix element for all element in a matrix.
@@ -238,7 +245,7 @@ def execute_for_all_elements(f, tensor, *args):
     else:  # Base case: tensor is not a list, apply f
         return f(tensor, *args)
 
-def clip(value, clip, no_overflow=False):
+def clip(value, clip, no_overflow=False, unsigned=False):
     """Value to possibly clip is clipped to max of bit length set by clip
     params:
     value = value to clip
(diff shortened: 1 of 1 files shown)


vp/devel/python_peripherals/DLA.py
```diff
@@ -248,15 +255,22 @@ def clip(value, clip, no_overflow=False):
...
-        if no_overflow:
-            value
-        return (value, 0)
+        if value > mask:
+                return mask-1 if no_overflow else (mask-1, value - mask)
+        elif value < -mask:
+                return -mask if no_overflow else (-mask, value + mask)
+
+    return value if no_overflow else (value, 0)
+
```
hegza: The control flow of this function gets a bit hard to follow. Consider splitting to `clip_signed` and `clip_unsigned`.

hegza: Good rule of thumb: if you're going to your third level of indentation, reconsider what you're doing.

vp/devel/python_peripherals/DLA.py
```diff
@@ -248,15 +255,22 @@ def clip(value, clip, no_overflow=False):
...
+            return mask if no_overflow else (mask, value - mask)
+        elif value < 0:
+            return 0 if no_overflow else (0, value)
+
     else:
-        if no_overflow:
-            value
-        return (value, 0)
+        if value > mask:
+                return mask-1 if no_overflow else (mask-1, value - mask)
```
hegza: Yo, inconsistent indentation.

vp/devel/python_peripherals/DLA.py
```diff
@@ -300,30 +305,26 @@ def write(self, data):
...
+        data -- [Int] data to write to bank
 
-        Return:
-        free -- Int Free space in bank
+        Returns:
+        unwritten -- [Int] data that didn't fit to bank
         """
-        return self.size - self.idx
+        assert(offset < self.size)
+        self.mem[offset] = data & 0xFF # Enforce 8bit width
```
hegza: Nice doc, I can read this.

However, I think it's not entirely accurate in this case. Looks stale :)

hegza: To be clear, this function does not return anything, unlike what is claimed.

vp/devel/python_peripherals/DLA.py
```diff
@@ -300,30 +305,26 @@ def write(self, data):
...
+
 
+    def read(self, offset):
```
hegza, line 331: By the way should these have some more expressive name such as `mmio_read` or `dla_read`? I think the `read` is a bit too generic for how these are used.

vp/devel/python_peripherals/DLA.py
```diff
@@ -435,68 +451,101 @@ def handle_bank_write(self, request):
...
-                    slice = data[i:]
-                    i = len(data)
-                else:
-                    slice = data[i:i+bank.available_memory()]
-                    i = i + bank.available_memory()
-                bank.write(slice)
-        if i != len(data):
-            return data[i:]
-        return []
+                print("Output written outside memory region")
```
hegza: Is this an error, warn or trace? Consider tagging it as appropriate, e.g. `print("ERROR: output written outside memory region)`.

vp/devel/python_peripherals/DLA.py
```diff
@@ -604,18 +658,34 @@ def pp_clip(self, values):
         """Clip pp values if register is set"""
         clip_amount = self.get_register(PP_CTRL, PP_CLIP_OFFSET, 5)
         if clip_amount > 0:
-            return execute_for_all_elements(clip, values, clip_amount, True)
+            return execute_for_all_elements(clip, values, clip_amount, True, True)
```
hegza: Pretty cool, this higher order function ':D

hegza: I wonder if there's a map operation for Python 3.4. It might be more readable.

hegza (changes requested):
Minor changes.

I'm assuming behavior is tested correct. We might want to review this more holistically at some point but I'm thinking it looks pretty good, considering available tooling.

vp/devel/python_peripherals/DLA.py
vilukissa68: Atleast that particular function is a method in the MemoryBank class. So the usage is always `bankobj.read(addr)`. Personally I think that this makes more sense than `bankobj.bank_read(addr)`, since there is only one kind of memory one can read from the bank.

vp/devel/python_peripherals/DLA.py
vilukissa68: It might have been that 3.4 Map() couldn't handle nested lists easily. I did try using it, but went with this fun programming exercise instead.

vp/devel/python_peripherals/DLA.py
vilukissa68: That's a good rule! Function splitted.

vp/devel/python_peripherals/DLA.py
vilukissa68: Don't know what happened here, probably some emacs shenanigans. Fixed indentation.

vp/devel/python_peripherals/DLA.py
vilukissa68: Documentation fixed

vp/devel/python_peripherals/DLA.py
vilukissa68: Tagged as a warning. This is indented functionality on the DLA but is not possible in VP due to it not having free access to the SDRAM.

vp/devel/python_peripherals/DLA.py
hegza: True, sounds good to me.

hegza (approved):
I think it's good enough. Let's do a more holistic review of the VP later.

PR #23, opened 2024-05-06
Encaptulates dla in struct. Adds Vec operations.

vilukissa68:
* Having dla in struct allows driver to remember setup between layers. This will be necessary for more complex layers.
* Vec<u8> enables run time allocation for dla layers, which allows for better flexibility.
* Adds example of usage in mac_benchmark example.
* Some minor features added
* Fixes lots of minor errors in driver

Depends on #22.


examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,93 @@
 #![no_std]
 
-mod mmap;
+extern crate alloc;
```
hegza, line 3: `extern crate` should not be necessary on 2021 edition, as alloc is imported implicitly upon first use. I recommend removing this line.

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
+#![no_std]
+#![no_main]
+
+extern crate alloc;
```
hegza, line 4: same here

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,93 @@
...
+mod mmap;
+pub use mmap::{
+    DLA0_ADDR, MEMORY_BANK_0_OFFSET, MEMORY_BANK_10_OFFSET, MEMORY_BANK_11_OFFSET,
+    MEMORY_BANK_12_OFFSET, MEMORY_BANK_13_OFFSET, MEMORY_BANK_14_OFFSET, MEMORY_BANK_15_OFFSET,
+    MEMORY_BANK_1_OFFSET, MEMORY_BANK_2_OFFSET, MEMORY_BANK_3_OFFSET, MEMORY_BANK_4_OFFSET,
+    MEMORY_BANK_5_OFFSET, MEMORY_BANK_6_OFFSET, MEMORY_BANK_7_OFFSET, MEMORY_BANK_8_OFFSET,
+    MEMORY_BANK_9_OFFSET, MEMORY_BANK_BASE_ADDR,
+};
+
+use alloc::vec::*;
```
hegza: I'm actually quite interested in what gets imported here as it is critical to understanding the allocation model of the code. I would prefer to avoid the wild card here and write open explicitly the imports.

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+
+use dla_driver::*;
+use headsail_bsp::{init_alloc, rt::entry, sprint, sprintln};
+use panic_halt as _;
+
+use rand::rngs::SmallRng;
+use rand::RngCore;
+use rand::{Rng, SeedableRng};
+
+use alloc::vec::*;
```
hegza: same here

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
+fn u4_to_u8(value: u8) -> u8 {
+    return value & 0xF;
 }
 
-pub fn dla_read(buf: &mut [u8], len: usize, offset: usize) {
-    for i in 0..len {
-        unsafe { buf[i] = ptr::read_volatile((DLA0_ADDR + offset + i) as *mut u8) }
+impl Dla {
+    pub fn new() -> Self {
+        return Dla {
```
hegza: `return` statement unnecessary and unidiomatic. Please use return expression instead e.g., just `Dla { .. }` not `return DLA { .. };`

examples/hpc/dla-driver/examples/dla.rs
```diff
@@ -8,8 +8,9 @@ use panic_halt as _;
 #[entry]
 fn main() -> ! {
     sprintln!("Hello world!");
-    dla_write_str("Hello DLA");
-    dla_init();
+    let mut dla = Dla::new();
+    dla.write_str("Hello DLA");
+    dla.init();
```
hegza: Is there ever a reason for needing `Dla` in an uninitialized state? If not, I would merge the `Dla::init(&self)`  into `Dla::new() -> Dla`. See also "resource acquisition is initialization (RAII)".

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+    (($input_h:expr, $kernel_h:expr, $padding_h:expr, $dilation_h:expr, $stride_h:expr)) => {{
+        ($input_h + 2 * $padding_h - $dilation_h * ($kernel_h - 1) - 1) / $stride_h + 1
+    }};
+}
+
+macro_rules! conv2d_out_parameters_width {
+    (($input_w:expr, $kernel_w:expr, $padding_w:expr, $dilation_w:expr, $stride_w:expr)) => {{
+        ($input_w + 2 * $padding_w - $dilation_w * ($kernel_w - 1) - 1) / $stride_w + 1
+    }};
+}
```
hegza: Seems like this could be achieved with a plain old function :- ) Could you justify the use of a macro here?

hegza: These are also apparently not used. Perhaps remove then?

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+
+fn conv2d_output_parameters(
```
hegza: The purpose of the function is not clear from the name. Consider using a verb, e.g., something along the lines of `calculate_conv2d_out_param_dim`.

examples/hpc/dla-driver/examples/mac_benchmark.rs
hegza, line 1: We seem to do lots of work with 2D arrays here. Consider declaring a type separate from `Vec` for these. E.g.,

```rust
struct Vec2<T> {
    elems: Vec<T>,
    width: usize,
    height: usize,
}
```

This can help with tracking the width & height when moving the Vec from one function to another.

If you want to go all-in with full-featured, optimized matrix types, I recommend [`ndarray`](https://crates.io/crates/ndarray). It has the same API as NumPy arrays.

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+fn generate_random_matrix(height: usize, width: usize, seed: u64) -> Vec<u8> {
+    let mut res: Vec<u8> = Vec::new();
+    let mut rng = SmallRng::seed_from_u64(seed);
+    for i in 0..(height * width) {
+        res.push((rng.next_u64() & 0xFF) as u8);
+    }
+    res
+}
```
hegza: `i` is not used. Replace with wildcard `_`. Consider using clippy to detect these.

There's a fine functional way to express all these if you're interested but I don't mind either solution:

```rust
let rng = SmallRng::seed_from_u64(seed);
(0..height * width).map(|_| (rng.next_u64() & 0xff) as u8).collect()
```

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+
+fn generate_random_matrix_small(height: usize, width: usize, seed: u64) -> Vec<u8> {
+    let mut res: Vec<u8> = Vec::new();
+    let mut rng = SmallRng::seed_from_u64(seed);
+    for i in 0..(height * width) {
```
hegza: same as above

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+    res
+}
+
+fn run_random_layer(
+    dla: &mut Dla,
+    in_w: usize,
+    in_h: usize,
+    k_w: usize,
+    k_h: usize,
```
hegza: Better names (preferred) or doc like

```rust
/// # Arguments
///
/// * `k_w` - kernel width
/// * `k_h` - kernel height
```

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,93 @@
...
+    BANK7,
+    BANK8,
+    BANK9,
+    BANK10,
+    BANK11,
+    BANK12,
+    BANK13,
+    BANK14,
+    BANK15,
+}
```
hegza: Extra credit: I'd format these along max char width rule rather than the default rule of one item per line. There's a an attribute you can use to disable rustfmt for a specific item and format it yourself.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,93 @@
...
+    BANK6,
+    BANK7,
+    BANK8,
+    BANK9,
+    BANK10,
+    BANK11,
+    BANK12,
+    BANK13,
+    BANK14,
+    BANK15,
```
hegza: Style: `Bank0` instead of `BANK0`.

I'm starting to feel like adding clippy to our CI pipeline... :)

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
 
-pub fn dla_write_str(s: &str) {
-    for b in s.as_bytes() {
-        unsafe { ptr::write_volatile(DLA0_ADDR as *mut u8, *b) };
-    }
-}
-pub fn dla_write(offset: usize, value: u8) {
-    unsafe { ptr::write_volatile((offset) as *mut u8, value) };
+macro_rules! get_bits {
+    ($mask:expr, $reg:expr) => {
```
hegza: Parameter order is strange. I'd put the $reg first any time of day.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
-    unsafe { ptr::write_volatile((DLA0_ADDR + offset) as *mut u32, value) }
+fn u2_to_u8(value: u8) -> u8 {
+    return value & 0x3;
 }
 
-pub fn dla_read_reg(offset: usize) -> u32 {
-    unsafe { ptr::read_volatile((DLA0_ADDR + offset) as *mut u32) }
+fn u4_to_u8(value: u8) -> u8 {
+    return value & 0xF;
 }
```
hegza: These functions do little to improve code readability :sweat_smile: I'd prefer having the bit-masks in straight line code, as the name `u2_to_u8` seems to mostly cause confusion.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
-        DLA_BUF_DATA_BANK_B_OFFSET,
-        DLA_BUF_DATA_BANK_B_BITMASK,
-        reg,
-        bank
-    );
-    dla_write_reg(DLA_BUF_DATA_BANK, reg);
-}
+    pub fn read_reg(&self, offset: usize) -> u32 {
+        unsafe { ptr::read_volatile((DLA0_ADDR + offset) as *mut u32) }
+    }
```
hegza: Use consistent naming, e.g.: `write_u8`, `write_u32`, `read_u32`.

In some cases it would be fine to do a generic impl e.g., `read<T>(&self, offset: usize) -> T` but I think the explicit `_usomething` is best here.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
+            let mut result: u128 = 0;
+            for i in 0..4 {
+                result |= (unsafe {
+                    ptr::read_volatile((MEMORY_BANK_BASE_ADDR + offset + (i * 4)) as *mut u32)
+                } as u128)
+                    << (32 * i)
+            }
+            result
+        } else {
+            unsafe { ptr::read_volatile((MEMORY_BANK_BASE_ADDR + (offset & !0xF)) as *mut u128) }
```
hegza: What's going on here with the mystery 8-bit bitmask? :)

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
-    let mut reg = dla_read_reg(DLA_BUF_CTRL);
-    reg = set_bits!(
-        DLA_READ_B_VALID_OFFSET,
-        DLA_READ_B_VALID_BITMASK,
-        reg,
-        ready as usize
-    );
-    dla_write_reg(DLA_BUF_CTRL, reg);
-}
+    pub fn read_data_bank(&self, offset: usize, len: usize) -> Vec<u8> {
```
hegza: Does this need to be public? It seems to be referenced only by DLA driver functions.

hegza: There is something suspicious about this API. It seems like it would allow arbitrary memory reads. Would it be possible to constrain its capabilities by using the `Bank` enum as a parameter instead of the `offset`?

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,93 @@
...
+};
+
+use alloc::vec::*;
 use core::ptr;
-use headsail_bsp::sprint;
+use headsail_bsp::{sprint, sprintln};
 use mmap::*;
 
+pub struct Dla {
+    simd_mode: SimdBitMode,
```
hegza: Is it necessary to cache this value? Can the hardware do it for us? Reduce runtime state if possible. This is an easy way to make hard to track implementation bugs less likely.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,93 @@
...
+    input_bank: MemoryBank,
+    kernel_bank: MemoryBank,
+    output_addr: usize,
```
hegza: Caching thing may apply here as well.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +100,493 @@ macro_rules! set_bits {
...
+        );
+        self.write_reg(DLA_PP_CTRL, reg);
+    }
+
+    pub fn is_ready(&self) -> bool {
+        let status = self.read_reg(DLA_STATUS_ADDR);
+        return !get_bits!(DLA_BUF_DONE_BITMASK, status) != 0;
+    }
+
+    pub fn set_bias_addr(&self, addr: u32) {
```
hegza, line 488: Address is `u32`. Is this correct or should it be platform-specific `usize`? I.e., can it be 64-bits when dispatched by HPC?

hegza: This is an interesting case study. We are indeed recording an _address_ in a register using a concrete, specific value. The right type here depends on the actual hardware capability, I think that should be `u64`.

examples/hpc/dla-driver/src/mmap.rs
```diff
@@ -71,17 +72,17 @@ pub(crate) const DLA_CTRL_VLD_BITMASK_UNUSED: usize = 0b10000000000;
 pub(crate) const DLA_PP_CLIP_BITMASK: usize = 0b111110000000000000000;
 
 pub(crate) const DLA_BUF_INPUT: usize = 0x14;
-pub(crate) const DLA_BUF_INPUT_WIDTH_OFFSET: usize = 0x0;
-pub(crate) const DLA_BUF_INPUT_HEIGHT_OFFSET: usize = 0x9;
-pub(crate) const DLA_BUF_INPUT_CHANNELS_OFFSET: usize = 0x18;
```
hegza, line 76: Double check these changes are correct. We are changing the offset from `0x18` to `18` which means that the value changes from 24 to 18.

examples/hpc/dla-driver/src/mmap.rs
```diff
@@ -116,9 +117,9 @@ pub(crate) const DLA_PP_INPUT_HEIGHT_BITMASK: usize = 0b111111111000000000000000
 
 pub(crate) const DLA_BUF_DATA_BANK: usize = 0x2C;
 pub(crate) const DLA_BUF_DATA_BANK_A_OFFSET: usize = 0x0;
-pub(crate) const DLA_BUF_DATA_BANK_B_OFFSET: usize = 0x10;
+pub(crate) const DLA_BUF_DATA_BANK_B_OFFSET: usize = 16;
 pub(crate) const DLA_BUF_DATA_BANK_A_BITMASK: usize = 0b1111;
-pub(crate) const DLA_BUF_DATA_BANK_B_BITMASK: usize = 0b11110000;
+pub(crate) const DLA_BUF_DATA_BANK_B_BITMASK: usize = 0b11110000000000000000;
```
hegza: I think `0b1111 << 16` is going to be more readable here.

hegza (changes requested):
Looks OK with minor changes. The PR is huge though, I'd like to have a separate live review of the entire driver. I'll try to schedule something.

examples/hpc/dla-driver/examples/dla.rs
vilukissa68: This particular instance of `dla.init()` is actually a typo, since there is no method called `dla_init()` anymore. It should read `dla.init_layer()` instead. As I understand from the DLA documentation this is a mandatory step of configuring the DLA before proceeding to writing the data for the next layer. So in that sense it is not initializing DLA, even though this a necessary step before actually doing calculations. Since the function needs to be called for each layer I thought that having to explicitly call it after acquisition made sense.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: The bitmask makes sure that the buffer access is aligned to 128-bits.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: The u32 is correct since DLA uses 32-bit registers. This method sets the memory location for bias data to DLA, so it's an actual value that the DLA interprets as a memory location.

examples/hpc/dla-driver/src/mmap.rs
vilukissa68: This is in fact correct. The hexadecimal value was a typo that started causing issues only now.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: I did try to remove this line but I get the following error
```error[E0433]: failed to resolve: use of undeclared crate or module alloc
  --> dla-driver/src/lib.rs:12:5
   |
12 | use alloc::vec::Vec;
   |     ^^^^^ use of undeclared crate or module alloc
   |
   = help: add `extern crate alloc` to use the alloc crate

```
Ironically the solution is to add back the extern. I don't think that I understand the rust module system enough yet to debug the issue.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Same issue here

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Removed wildcard, using explicit imports.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Removed wildcard, using explicit imports.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: return removed.

examples/hpc/dla-driver/examples/dla.rs
vilukissa68: Init not merged to `Dla::new()`

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: These are leftovers from before we had dynamic allocation. Removed.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Named to `calculate_conv2d_out_param_dim`

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: I'll put this for consideration for future

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Unused variable replaced with wildcard. Imperative syntax kept.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Unused variable replaced with wildcard. Imperative syntax kept.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Short forms opened, to be more descriptive.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: rustfmt overridden for this enum with `#[rustfmt::skip]` and formatting done manually.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Emacs default linter configuration doesn't complain about upper case enums. Changed to use preferred style.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Swapped parameter order

examples/hpc/dla-driver/src/lib.rs
vilukissa68: This is the consequence of trying to abstract functionality before figuring out that the solution is too simple to abstract 😅. Removed these.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Renamed register methods to `write_u32` and `read_u32`.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: So the asic DLA can write results to any arbitrary memory address in SDRAM, thus the DLA driver needs to be able to do the same. The method `read_data_bank()` is thus poorly named. I implemented the restriction for this method to only allow memory access for the banks, but in the future dla driver needs to be able to read the result from arbitrary memory locations.

examples/hpc/dla-driver/src/mmap.rs
vilukissa68: Yeah this is way better. Modified longer bitmasks to use this syntax.

examples/hpc/dla-driver/src/lib.rs
hegza: Oh, right. I think alloc is special cased. As is compiler_builtins. My bad :D

examples/hpc/dla-driver/examples/dla.rs
hegza: I think the per-layer API is best as discussed over lunch. This is ok for now; we can change it later.

examples/hpc/dla-driver/src/lib.rs
hegza: This is sus. I'd delegate this to a user problem. Just assert the alignment instead of editing the input.

vilukissa68: Removed

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Unnecessary caching removed and replaced with get calls to hardware.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Unnecessary caching removed and replaced with get calls to hardware

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +106,501 @@ macro_rules! set_bits {
...
+impl Dla {
+    pub fn new() -> Self {
+        Dla {
+        }
+    }
+    pub fn write_str(&self, s: &str) {
+        for b in s.as_bytes() {
+            unsafe { ptr::write_volatile(DLA0_ADDR as *mut u8, *b) };
+        }
+    }
```
hegza: Wait a second, what is this function achieving? It writes a string at the start of DLA0, is there a UART here or was this intended as a debug thing? Should this be removed?

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +106,501 @@ macro_rules! set_bits {
...
+    pub fn write(&self, offset: usize, value: u8) {
```
hegza: Consider `write_u8` for consistency.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +106,501 @@ macro_rules! set_bits {
...
+    pub fn read_u32(&self, offset: usize) -> u32 {
+        unsafe { ptr::read_volatile((DLA0_ADDR + offset) as *mut u32) }
     }
-}
 
-pub fn dla_write_data_bank(offset: usize, buf: &mut [u8]) {
-    sprint!("\nWrite to bank {:#x}, data: {:?}", offset, buf);
-    for (i, b) in buf.iter().enumerate() {
-        unsafe { ptr::write_volatile((MEMORY_BANK_BASE_ADDR + offset + i) as *mut u8, *b) };
+    pub fn read(&self, buf: &mut [u8], len: usize, offset: usize) {
```
hegza: Recommended parameter ordering: `offset, len, &mut buf`

hegza: Recommended fn name: `read_bytes`

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -1,11 +1,99 @@
...
 use core::ptr;
-use headsail_bsp::sprint;
+use headsail_bsp::{sprint, sprintln};
 use mmap::*;
 
+pub struct Dla {
+}
+
+#[rustfmt::skip]
+pub enum MemoryBank {
```
hegza, line 24: Consider making this Clone & Copy: `#[derive(Clone, Copy)]`. Integer sized types are cheap to copy and you won't have to bother with the borrow checker (i.e., `.clone` and ownership issues) in the user code.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -18,291 +106,501 @@ macro_rules! set_bits {
...
+        MemoryBank::from_u32(reg)
+    }
+
+    pub fn get_output_bank(&self) -> MemoryBank {
+        let reg = self.read_u32(DLA_PP_AXI_WRITE);
+        let bank_idx : u32 = (reg - MEMORY_BANK_BASE_ADDR as u32) / MEMORY_BANK_SIZE as u32;
+        MemoryBank::from_u32(bank_idx)
+    }
+
+    pub fn set_mac_clip(&self, clip_amount: usize) {
```
hegza: Should this be `u32`? `usize` seems unlikely.

hegza (changes requested):
Some changes that should be helpful when moving along.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: This was a leftover from the first version of dla driver, where it was used to write hello world to memory. Removed.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Renamed as suggested

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Renamed function and reordered parameters as suggested.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Added Clone & Copy, but not yet used anywhere.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Mac clip is 5 bits wide, so I don't really know how much does `usize` vs `u32` matter here. Changed to `u32`.

examples/hpc/dla-driver/src/lib.rs
hegza: You'll notice it when you do something like this:

```rust
let dla = Dla::new();

let sel_bank = MemoryBank::Bank4;
dla.set_bank_attribute_1(sel_bank, attr);
dla.set_bank_attribute_2(sel_bank, attr2);
dla.set_bank_attribute_3(sel_bank, attr3);
```

Without `derive(Copy)` you'd need to `.clone()` on every `sel_bank` above. Without `derive(Clone)`, the above code wouldn't work at all but you'd have to `let sel_bank = ...` a new instance for every function call.

examples/hpc/dla-driver/src/lib.rs
hegza: `usize` is an architecture dependent type and it immediately raises a question about "why". Better to use a concrete well-defined type.

hegza (approved):
All good. Consider rebasing on latest main before merge.

hegza (approved):

PR #24, opened 2024-05-08
Feat/weaken symbols ffi

vilukissa68:
Generates headsail-bsp-ffi with weak symbols to enable use with libgcc

### DIFF
--- a/examples/headsail-bsp-ffi/Justfile
+++ b/examples/headsail-bsp-ffi/Justfile
@@ -3,4 +3,4 @@ build-sysctrl:
 
 build-hpc:
     cargo build --target=riscv64imac-unknown-none-elf -Fpanic-uart
-
+    sh weaken_compiler_builtins.sh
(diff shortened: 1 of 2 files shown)


examples/headsail-bsp-ffi/Justfile
```diff
@@ -1,6 +1,7 @@
 build-sysctrl:
     cargo build --target=riscv32imc-unknown-none-elf -Fpanic-uart
+    sh weaken_compiler_builtins.sh
```
hegza: Put a shebang as the first line of the file instead of piping to sh here.

`#!/bin/bash` or similar

hegza (approved):

examples/headsail-bsp-ffi/Justfile
vilukissa68: Shebang added and piping removed.

PR #25, opened 2024-05-13
Adds struct for configuring dla layers

vilukissa68:
Adds LayerConfig struct for interfacing with the driver. Layer configuration with struct ensures that the state of the hardware is explicitly set between layers. This made it possible to reduce the amount of public methods in the DLA API.


examples/hpc/dla-driver/src/lib.rs
```diff
@@ -16,7 +16,30 @@ use core::ptr;
 use headsail_bsp::{sprint, sprintln};
 use mmap::*;
 
-pub struct Dla {
+pub struct LayerConfig {
```
hegza, line 70: Consider making optional fields `Option<u32>` and grouping related attributes into tuples or structs.

hegza: e.g., `kernel_size: KernelSize { channels, width, height }`, and if it's optional: `kernel_size: Option<KernelSize>`

hegza (approved):
Mostly suggestions this time. Looks fine.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Good idea. This will make the API more user-friendly since I suspect that some options are somewhat rarely used. I added Option fields and grouped related attributes to structs. I also added sensible default values for option types.

hegza (approved):
`Option<bool>` seems superfluous. The difference between `None` and `false` does not justify the added layer of abstraction. I'd just leave them as `bool`.

I imagine this struct is used something like this:

```rust
let config = LayerConfig { bias_addr: Some(0x1000), .. /* Functional update syntax `..` takes rest using Default::default() */ };
dla.set_layer(config);
```

Used like this, the default value of `false` will be used for boolean parameters.

If only certain parameter combinations make sense, there might be a better yet API using the builder pattern e.g., `LayerConfig::default().bias(0x1000)` but I don't care too much for that. If it works its good enough for me.

vilukissa68:
Option removed from the booleans. I suppose that since previous changes were approved this is good to merge.

PR #27, opened 2024-05-14
Adds signed integers to DLA

vilukissa68:
* Adds support for 8-bit signed integers to driver and VP
* Removes support for unsigned integers (u8) since those turned out not to be supported by the hardware
* Fixes signed clipping bounds
* Fixes PP clipping being dependant on ReLU

### DIFF
--- a/examples/hpc/dla-driver/examples/mac_benchmark.rs
+++ b/examples/hpc/dla-driver/examples/mac_benchmark.rs
@@ -28,7 +28,7 @@ fn calculate_conv2d_out_param_dim(
 fn generate_random_array(buffer: &mut [i8], size: usize) {
     let mut rng = SmallRng::seed_from_u64(1234567890);
     for i in 0..size {
-        buffer[i] = (((rng.next_u64() % u8::MAX as u64) as i8) + i8::MIN) as i8;
+        buffer[i] = rng.next_u64() as i8;
     }
 }
 
@@ -36,7 +36,7 @@ fn generate_random_matrix(height: u32, width: u32, seed: u64) -> Vec<i8> {
     let mut res: Vec<i8> = Vec::new();
     let mut rng = SmallRng::seed_from_u64(seed);
     for _ in 0..(height * width) {
-        res.push((((rng.next_u64() % u8::MAX as u64) as i8) + i8::MIN) as i8);
+        res.push((rng.next_u64() & 0xFF) as i8);
     }
     res
 }
@@ -45,7 +45,7 @@ fn generate_random_matrix_small(height: u32, width: u32, seed: u64) -> Vec<i8> {
     let mut res: Vec<i8> = Vec::new();
     let mut rng = SmallRng::seed_from_u64(seed);
     for _ in 0..(height * width) {
-        res.push((((rng.next_u64() % 4) as i8) - 2) as i8);
+        res.push((rng.next_u64() & 0x3) as i8);
     }
     res
 }
--- a/examples/hpc/dla-driver/examples/mac_benchmark.rs
+++ b/examples/hpc/dla-driver/examples/mac_benchmark.rs
@@ -28,7 +28,7 @@ fn calculate_conv2d_out_param_dim(
 fn generate_random_array(buffer: &mut [i8], size: usize) {
     let mut rng = SmallRng::seed_from_u64(1234567890);
     for i in 0..size {
-        buffer[i] = rng.next_u64() as i8;
+        buffer[i] = (((rng.next_u64() % u8::MAX as u64) as i8) + i8::MIN) as i8;
     }
 }
 
@@ -36,7 +36,7 @@ fn generate_random_matrix(height: u32, width: u32, seed: u64) -> Vec<i8> {
     let mut res: Vec<i8> = Vec::new();
     let mut rng = SmallRng::seed_from_u64(seed);
     for _ in 0..(height * width) {
-        res.push((rng.next_u64() & 0xFF) as i8);
+        res.push((((rng.next_u64() % u8::MAX as u64) as i8) + i8::MIN) as i8);
     }
     res
 }
@@ -45,7 +45,7 @@ fn generate_random_matrix_small(height: u32, width: u32, seed: u64) -> Vec<i8> {
     let mut res: Vec<i8> = Vec::new();
     let mut rng = SmallRng::seed_from_u64(seed);
     for _ in 0..(height * width) {
-        res.push((rng.next_u64() & 0x3) as i8);
+        res.push((((rng.next_u64() % 4) as i8) - 2) as i8);
     }
     res
 }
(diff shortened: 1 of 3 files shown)


vp/devel/python_peripherals/DLA.py
```diff
@@ -193,6 +193,22 @@ def flat_to_CWH(data, channels, width, height):
                 i = i + 1
     return output
 
+def cast_long_to_signed_byte(value):
```
hegza, line 196: I recommend adding type ascriptions in the signature for param and for return.

By personal preference, I'd name this something like `u32_into_i8` but use whatever is clearest to you.

hegza: Especially, I'd generally avoid the ABI-dependent vocabulary such as `long` and `int`.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -199,10 +199,10 @@ impl Dla {
         }
     }
 
-    pub fn write_data_bank(&self, offset: usize, buf: &mut [u8]) {
+    pub fn write_data_bank(&self, offset: usize, buf: &mut [i8]) {
         //sprintln!("\nWrite to bank {:#x}, data: {:?}", offset, buf);
         for (i, b) in buf.iter().enumerate() {
-            unsafe { ptr::write_volatile((MEMORY_BANK_BASE_ADDR + offset + i) as *mut u8, *b) };
+            unsafe { ptr::write_volatile((MEMORY_BANK_BASE_ADDR + offset + i) as *mut i8, *b) };
```
hegza: You might want to elide the type in these assignments, if types are obvious from context, e.g. `write_volatile(ADDR as *mut _, value)`. Say, you have a function like `write_u8()`, there is no need to re-specify the type in the implementation. This might come down to preference at the end of the day.

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -25,27 +25,27 @@ fn calculate_conv2d_out_param_dim(
     (output_width, output_height)
 }
 
-fn generate_random_array(buffer: &mut [u8], size: usize) {
+fn generate_random_array(buffer: &mut [i8], size: usize) {
     let mut rng = SmallRng::seed_from_u64(1234567890);
     for i in 0..size {
-        buffer[i] = rng.next_u64() as u8;
+        buffer[i] = rng.next_u64() as i8;
```
hegza, line 31: This might be wrong. `rng.next_u64` will return a 64-bit unsigned integer. `as` keyword will truncate bits, leaving you with a `u8`, never producing negative numbers. The intent of this code should be made clear at least. You can do that by making the conversions explicit, e.g.,

```rust
// Number transformation
let signed_byte = (((rng.next_u64() % u8::MAX) as u8) + i8::MIN) as i8;

// Bitwise transmute
let signed_byte: i8 = mem::transmute(rng.next_u64() % u8::MAX) as u8;
```

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -25,27 +25,27 @@ fn calculate_conv2d_out_param_dim(
...
 }
 
-fn generate_random_matrix(height: u32, width: u32, seed: u64) -> Vec<u8> {
-    let mut res: Vec<u8> = Vec::new();
+fn generate_random_matrix(height: u32, width: u32, seed: u64) -> Vec<i8> {
+    let mut res: Vec<i8> = Vec::new();
     let mut rng = SmallRng::seed_from_u64(seed);
     for _ in 0..(height * width) {
-        res.push((rng.next_u64() & 0xFF) as u8);
+        res.push((rng.next_u64() & 0xFF) as i8);
```
hegza, line 39: Same here, this will not be what you want, I think.

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -25,27 +25,27 @@ fn calculate_conv2d_out_param_dim(
...
 }
 
-fn generate_random_matrix_small(height: u32, width: u32, seed: u64) -> Vec<u8> {
-    let mut res: Vec<u8> = Vec::new();
+fn generate_random_matrix_small(height: u32, width: u32, seed: u64) -> Vec<i8> {
+    let mut res: Vec<i8> = Vec::new();
     let mut rng = SmallRng::seed_from_u64(seed);
     for _ in 0..(height * width) {
-        res.push((rng.next_u64() & 0x1) as u8);
+        res.push((rng.next_u64() & 0x3) as i8);
```
hegza, line 48: See above.

examples/hpc/dla-driver/examples/mac_benchmark.rs
```diff
@@ -116,7 +116,7 @@ fn run_random_layer(
     sprintln!("Waiting for calculation");
     while !dla.handle_handshake() {}
     sprintln!("Calculation ready");
-    let output: Vec<u8> = dla.read_output(output_width as usize * output_height as usize);
+    let output: Vec<i8> = dla.read_output(output_width as usize * output_height as usize);
     output
```
hegza: Style: return expression `dla.read_output(output_width as usize * output_height as usize)` without the extra assignment.

hegza (changes requested):
Possibly incorrect random generation going on.

vp/devel/python_peripherals/DLA.py
vilukissa68: So as far as I understand IronPython documentation, long doesn't equate to u32, but rather Python's infinite integer signified with L (e.g 42L). i8 makes more sense here than signed byte though.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Based on my experimentation the original code does in fact produce negative numbers. This is clear from the fact that the results for the MAC step in DLA produces negative values, which would not be possible if the input was all positive. The explicit conversion might be better, but for some reason, Renode hangs with it so I'm leaning towards keeping the original for now. We also have test inputs available from the DLA hardware testbench we should start transitioning soon anyway.

hegza: The hang is extremely suspicious 😃 There's probably something going on there that we don't understand. It's an indication that this code is doing something other than what we think it's doing.

I guess we can live with just knowing it works since it's just example generation and the results are correct.

examples/hpc/dla-driver/examples/mac_benchmark.rs
vilukissa68: Fixed this

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Removed explicit type assignments.

hegza (approved):
I think it's all good here.

PR #29, opened 2024-05-16
Add C test cases to CI

hegza:
There's currently an issue with the RISC-V GCC compiler on the CI runner.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,7 +16,7 @@ env:
   DLA_VALIDATION_BIN: validate
 
   # Used for setting up RISC-V GCC toolchain for C tests
-  RISCV: /tmp/riscv/
+  RISCV: /opt/riscv/
   RISCV_XLEN: riscv64
   GCC_VER: "2024.04.12"
   UBUNTU_VER: ubuntu-22.04
@@ -111,11 +111,6 @@ jobs:
   build-dla-validation:
     runs-on: ubuntu-latest
 
-    env:
-      RISCV: /opt/riscv/
-      GCC_VER: "2024.04.12"
-      RISCV_XLEN: riscv64
-      UBUNTU_VER: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
     - name: Create RISC-V GCC build directory
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -182,6 +182,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - name: Install requirements
+      run: cargo install just
     - name: Build minimal C project
       working-directory: ./examples/hpc-c/hello-hpc
       run: just build
@@ -197,6 +199,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - name: Install requirements
+      run: cargo install just
     - name: Build minimal C project
       working-directory: ./examples/hpc-c/test-bsp-ffi
       run: just build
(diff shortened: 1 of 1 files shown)


vilukissa68:
Got this working in #46. There were couple issues:
1. URL for the toolchain was wrong with `ubuntu-` being duplicated.
2. `ubuntu-latest` seems to have an issue with permission in `/root` dir. This was solved by unpacking the toolchain in `/tmp`
3. Adding to `$PATH` doesn't work with GitHub actions, since environment variables get cleaned between steps. This is solved by appending the toolchain to `$GITHUB_PATH`

In the end steps for adding the toolchain look something like this.

```yaml
    - name: Create RISC-V GCC build directory
      run: mkdir /tmp/riscv-gnu-toolchain
    - name: Download GCC
      working-directory: /tmp/riscv-gnu-toolchain/
      run: |
        wget -q https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/$GCC_VER/$RISCV_XLEN-elf-$UBUNTU_VER-gcc-nightly-$GCC_VER-nightly.tar.gz
        mkdir ${RISCV} && \
          tar -xf $RISCV_XLEN-elf-$UBUNTU_VER-gcc-nightly-$GCC_VER-nightly.tar.gz --strip-components 1 -C $RISCV
        export PATH="${RISCV}/bin:${PATH}"
        echo "${RISCV}/bin" >> $GITHUB_PATH
        $RISCV_XLEN-unknown-elf-gcc --version
        $RISCV_XLEN-unknown-elf-ld --version
```
Disclaimer this uses GCC instead of LLVM

hegza:
Based on latest runs it seems like the downloaded version of RISC-V is configured to use a different float setup than what's used by our build. Builds fine locally, since we use the multilib compiler.

It's not feasible to rebuild the compiler for each CI. Perhaps we could try build-minimal-c with newlib..? (@andstepan) Any other ideas? (@vilukissa68)

The end goal is to try to have C builds in CI.

vilukissa68:
> Based on latest runs it seems like the downloaded version of RISC-V is configured to use a different float setup than what's used by our build. Builds fine locally, since we use the multilib compiler.
> 
> It's not feasible to rebuild the compiler for each CI. Perhaps we could try build-minimal-c with newlib..? (@andstepan) Any other ideas? (@vilukissa68)
> 
> The end goal is to try to have C builds in CI.

I fail understand why this issue happens. It looks like the `ld` from the same toolchain as the `libc` conflict? Doesn't riscv-collab provide gcc with multilib?

We could build Headsail-newlib in the gitlab CI and pull that for C builds, since there is really only one way we want to configure it for headsail, but there's nothing saying that it will fix this issue if the compiler is not good.

If we can't find a suitable `gcc`, we could have an additional workflow which compiles `gcc` the way we want. And store that as artifact, so that main flow can just pull it. I would guess that it's possible to have flows in Github/Gitlab that are run only when manually evoked.

hegza:
> We could build Headsail-newlib in the gitlab CI and pull that for C builds, since there is really only one way we want to configure it for headsail, but there's nothing saying that it will fix this issue if the compiler is not good.

I like this. We'll need to coordinate some keys between the CIs to enable cross-downloads but that should be fine. I think it _should_ fix this.

> If we can't find a suitable `gcc`, we could have an additional workflow which compiles `gcc` the way we want. And store that as artifact, so that main flow can just pull it. I would guess that it's possible to have flows in Github/Gitlab that are run only when manually evoked.

This would work but building GCC takes literally hours. There's not too many minutes available to our org on the GitHub Free license :)

hegza:
Newlib artifacts now available at <https://gitlab.tuni.fi/soc-hub/headsail/sw/newlib-artifacts>. I will upload the necessary lib & headers in-tree and re-try this.

PR #30, opened 2024-05-17
Adds better conv2d implementation

vilukissa68:
* The previous conv2d only worked for 1-channel kernels. This has now been fixed, and conv2d outputs an equal amount of tensors as it has been given filters. Stride and padding functionality has also been confirmed to work as expected.
* Numpy style operations added to improve redability


vp/devel/python_peripherals/DLA.py
```diff
@@ -169,6 +169,101 @@ def reshape_to_cwh(data):
...
+
+    """
+    shape = []
+    while isinstance(tensor, list):
+        shape.append(len(tensor))
+        tensor = tensor[0]
+    return tuple(shape)
+
+def flatten(tensor):
+    """Flattens ndimensional tensor to one dimensional tensor/vector
```
hegza, line 226: In what order? Does it matter or is it obvious? Is there a method to invert this operation you could reference?

vp/devel/python_peripherals/DLA.py
```diff
@@ -576,31 +669,36 @@ def set_input_data(self, data):
...
-       while len(data) < (channels * width * height):
-           # Move to next bank
-           if offset > bank.size:
-               bank_idx = bank_idx + 1
-               bank = self.banks[bank_idx]
-               offset = 0
-           data.append(bank.read(offset))
-           offset += 1
-       return channels, width, height, data
+        channels -- Int Number of channels
```
hegza: The return value seems to have `filter_amount` as first element. Could this doc be stale or what's up?

vp/devel/python_peripherals/DLA.py
```diff
@@ -826,78 +902,92 @@ def conv2d_check_parameters(self, A, kernel, padding, dilation, stride):
         w_in = len(A[0])
         h_kernel = len(kernel)
         w_kernel = len(kernel[0])
-        h_out = math.floor((h_in + 2*padding[0] - dilation[0] * (h_kernel - 1) -1) / stride[0] +1)
+        h_out = math.floor(((h_in + 2*padding[0] - dilation[0] * (h_kernel - 1) -1) / stride[0]) +1)
```
hegza: This line has the extra parens but the below line doesn't?

hegza (approved):
I like it. Minor recommendations. No need to re-submit for review.

I think it's about time to start doing some unit testing (and CI) on the behavior to give some confidence that refactors don't break existing code.

vp/devel/python_peripherals/DLA.py
vilukissa68: Good point! I extended to support either row-major order and column-major order, with proper documentation.

vp/devel/python_peripherals/DLA.py
vilukissa68: Yes this was stale documentation. Fixed.

vp/devel/python_peripherals/DLA.py
vilukissa68: Extra parenthesis removed.

PR #31, opened 2024-05-17
Fixes style issues in DLA driver

vilukissa68:
* Adds some documentation to DLA rust driver
* Fixes multiple style issues mentioned in code review
* Implements TryFrom32 for MemoryBank

Most of the changes are probably good to go since no functional changes have been made, but I would like to get an opinion if the way I implemented the TryFrom for MemoryBank is sensible.

### DIFF
--- a/examples/hpc/dla-driver/src/lib.rs
+++ b/examples/hpc/dla-driver/src/lib.rs
@@ -1,3 +1,6 @@
+//! # DLA driver
+//!
+//! Implements driver for sochub headsail SoC's deep learning accelerator.
 #![no_std]
 
 extern crate alloc;
@@ -30,7 +33,7 @@ const DEFAULT_PADDING: PaddingConfig = PaddingConfig {
     right: 0,
     left: 0,
     bottom: 0,
-    value: 0,
+    padding_value: 0,
 };
 const DEFAULT_STRIDE: StrideConfig = StrideConfig { x: 1, y: 1 };
 const DEFAULT_MAC_CLIP: u32 = 8;
@@ -42,31 +45,54 @@ use core::ptr;
 use headsail_bsp::{sprint, sprintln};
 use mmap::*;
 
+/// Dimensions of kernel
 pub struct KernelSize {
     pub channels: u32,
     pub width: u32,
     pub height: u32,
 }
 
+/// Dimensions of inputs
 pub struct InputSize {
     pub channels: u32,
     pub width: u32,
     pub height: u32,
 }
 
+/// Conv2d padding
+/// Value used for padding the matrix
+///
+/// # Examples
+///
+/// Non functional example, padding is done in hardware
+/// ```
+/// let padding = PaddingConfig {top:1, right:1, left:1, bottom:1, padding_value: 7};
+/// matrix = [[1,2], [3,4]];
+/// pretty_print_matrix(matrix)
+/// 1 2
+/// 3 4
+/// padded_matrix = applyPadding(matrix, padding);
+/// pretty_print_matrix(padded_matrix);
+/// 7 7 7 7
+/// 7 1 2 7
+/// 7 3 4 7
+/// 7 7 7 7
+/// ```
 pub struct PaddingConfig {
     pub top: u32,
     pub right: u32,
     pub left: u32,
     pub bottom: u32,
-    pub value: u32,
+        pub padding_value: u32,
 }
 
+/// Conv2d stride
 pub struct StrideConfig {
     pub x: u32,
     pub y: u32,
 }
 
+/// Configures DLA for performing calculation for layers
 pub struct LayerConfig {
     pub input_bank: Option<MemoryBank>,
     pub kernel_bank: Option<MemoryBank>,
@@ -86,6 +112,7 @@ pub struct LayerConfig {
 
 #[derive(Clone, Copy)]
 #[rustfmt::skip]
+/// Data banks in DLA's memory buffer, stores inputs, kernels and outputs.
 pub enum MemoryBank {
     Bank0, Bank1, Bank2, Bank3, Bank4, Bank5, Bank6, Bank7, Bank8, Bank9,
     Bank10, Bank11, Bank12, Bank13, Bank14, Bank15,
@@ -157,6 +184,7 @@ impl MemoryBank {
 }
 
 #[derive(Copy, Clone)]
+/// DLA support three SIMD modes
 pub enum SimdBitMode {
     EightBits = 0,
     FourBits = 1,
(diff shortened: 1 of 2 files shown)


examples/hpc/dla-driver/src/lib.rs
```diff
@@ -42,31 +45,54 @@ use core::ptr;
...
+/// matrix = [[1,2], [3,4]];
+/// pretty_print_matrix(matrix)
+/// 1 2
+/// 3 4
+/// padded_matrix = applyPadding(matrix, padding);
+/// pretty_print_matrix(padded_matrix);
+/// 7 7 7 7
+/// 7 1 2 7
+/// 7 3 4 7
+/// 7 7 7 7
```
hegza, line 82: Nice.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -86,35 +112,39 @@ pub struct LayerConfig {
...
+/// Data banks in DLA's memory buffer, stores inputs, kernels and outputs.
 pub enum MemoryBank {
     Bank0, Bank1, Bank2, Bank3, Bank4, Bank5, Bank6, Bank7, Bank8, Bank9,
     Bank10, Bank11, Bank12, Bank13, Bank14, Bank15,
 }
 
-impl MemoryBank {
-    fn from_u32(value: u32) -> MemoryBank {
+impl TryFrom<u32> for MemoryBank {
+    type Error = &'static str;
```
hegza: If you're using a string, at least make it informative :-)

I'd use a unit type (`()`) or a trivial tuple struct here (`struct InvalidBankIdx(u32);`), since it should be obvious at the call site what caused the error. In the unlikely case the runtime/driver implementor wants to add more info, they can impl Debug for this struct.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -134,7 +164,7 @@ impl MemoryBank {
             MemoryBank::Bank15 => MEMORY_BANK_15_OFFSET,
         }
     }
-    fn value(&self) -> usize {
+    fn from(&self) -> usize {
```
hegza: A function like this should be called `into_usize` if implemented in `MemoryBank`. However, these conversions are usually implemented using `From`, `TryFrom`, `Into` and `TryInto`.

`From` and `TryFrom` are usually preferred over `*Into` because they provide the respective `Into` and `TryInto` impls automatically. Embedded code is an exception to this rule though, as people tend to be careful with code size. Another exception which applies here is the Rust orphan rule, i.e., if the target type is a foreign/primitive type, preventing `impl From<_> for <foreign>`.

**I think in this case the most appropriate choice is to implement `impl Into<usize> for MemoryBank {}`**.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -245,65 +272,74 @@ impl Dla {
...
 
+    /// Sets one of the DLA's memory banks as starting bank for inputs
     fn set_input_data_bank(&self, bank: MemoryBank) {
         let mut reg = self.read_u32(DLA_BUF_DATA_BANK);
         reg = set_bits!(
             DLA_BUF_DATA_BANK_B_OFFSET,
             DLA_BUF_DATA_BANK_B_BITMASK,
             reg,
-            bank.value()
+            bank.from()
```
hegza: Yeah this looks wrong :D see above. This should definitely be an `into`.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -483,46 +530,62 @@ impl Dla {
...
+    /// Reads index of the first input bank
     fn get_input_bank(&self) -> MemoryBank {
         let mut reg = self.read_u32(DLA_BUF_DATA_BANK);
         reg = get_bits!(reg, DLA_BUF_DATA_BANK_B_BITMASK);
-        MemoryBank::from_u32(reg)
+        let bank = match MemoryBank::try_from(reg) {
+            Ok(bank) => bank,
+            Err(e) => panic!("{}", e),
+        };
+        bank
```
hegza: If there's a way to assume that the register always contains a valid value, you should  use `.unwrap()` here. Say, we assume the hardware doesn't ever change the value and that we make sure to only store valid values with assertion / match on the store side.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -483,46 +530,62 @@ impl Dla {
...
+    /// Reads index of the first kernel bank
     fn get_kernel_bank(&self) -> MemoryBank {
         let mut reg = self.read_u32(DLA_BUF_DATA_BANK);
         reg = get_bits!(reg, DLA_BUF_DATA_BANK_A_BITMASK);
-        MemoryBank::from_u32(reg)
+        let bank = match MemoryBank::try_from(reg) {
+            Ok(bank) => bank,
+            Err(e) => panic!("{}", e),
+        };
+        bank
```
hegza: Same as above.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -483,46 +530,62 @@ impl Dla {
...
+    /// Reads index of the first output bank
     fn get_output_bank(&self) -> MemoryBank {
         let reg = self.read_u32(DLA_PP_AXI_WRITE);
         let bank_idx: u32 = (reg - MEMORY_BANK_BASE_ADDR as u32) / MEMORY_BANK_SIZE as u32;
-        MemoryBank::from_u32(bank_idx)
+        let bank = match MemoryBank::try_from(bank_idx) {
+            Ok(bank) => bank,
+            Err(e) => panic!("{}", e),
+        };
+        bank
```
hegza: same

examples/hpc/dla-driver/src/lib.rs
hegza: I know from experience that `derive(Debug)` can easily become an issue on code size constrained implementations (such as anything on SysCtrl). What I'd do nowadays would be to add an optional `derive(Debug)` on `InvalidBankIdx`, either by using `cfg_attr(debug_assertions, derive(Debug))` or providing a feature such as `debug-traits`.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -483,46 +530,62 @@ impl Dla {
...
 
+    /// Sets clipping after post-processing
     fn set_pp_clip(&self, clip_amount: u32) {
-        let mut reg = self.read_u32(DLA_PP_CTRL);
         // Cap clipping amount
         if clip_amount > 0x1F {
-            reg = set_bits!(DLA_PP_CLIP_OFFSET, DLA_PP_CLIP_BITMASK, reg, 0x1F);
-        } else {
-            reg = set_bits!(DLA_PP_CLIP_OFFSET, DLA_PP_CLIP_BITMASK, reg, clip_amount);
+            panic!("Clip can't be more than 20")
```
hegza: Consider returning `Result<(), InvalidClip(u32)>`, as strings and panics are not always the most convenient choice for debugging at this level. This way the correct display format can be chosen at the user level.

hegza (changes requested):
Good try! Some corrections to error handling and conventional impls.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Changed to use unit type.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: `impl Into<usize> for MemoryBank {}` implemented and used

examples/hpc/dla-driver/src/lib.rs
vilukissa68: changed to into

examples/hpc/dla-driver/src/lib.rs
vilukissa68: changed to use `.unwrap()`

examples/hpc/dla-driver/src/lib.rs
vilukissa68: changed to use `.unwrap()`

examples/hpc/dla-driver/src/lib.rs
vilukissa68: changed to use `.unwrap()`

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Both clip setting functions now return `Result<(), InvalidClip>`

hegza (approved):

PR #32, opened 2024-05-17
Adds relocatable timer driver for HPC SS

andstepan:
This pull request adds relocatable drivers for the timers of the HPC subsystem. The drivers have been implemented as structs with a generic constant.
Moreover, the base addresses for timers 1 to 3 have been added to the mmap.rs file.
Finally, new documentation has been added and existing documentation has been rewritten in a Rust doc-friendly way.

### DIFF
--- a/examples/headsail-bsp/src/apb_timer.rs
+++ b/examples/headsail-bsp/src/apb_timer.rs
@@ -44,7 +44,7 @@ impl<const BASE_ADDRESS: usize> ApbTimer<BASE_ADDRESS>{
         // Read register
         let mut reg = read_u32(BASE_ADDRESS + Self::TIMER_CTRL_REG_OFFSET);
         // Write 0 to bit 0 but leave all other bits untouched
-        reg.set_bit(Self::TIMER_ENABLE_BIT as usize, false);
+        reg.set_bit(Self::TIMER_ENABLE_BIT, false);
         // Write register back
         write_u32(BASE_ADDRESS + Self::TIMER_CTRL_REG_OFFSET, reg);
     }
--- a/examples/headsail-bsp/src/apb_timer.rs
+++ b/examples/headsail-bsp/src/apb_timer.rs
@@ -1,4 +1,4 @@
-/**
+/*!
  * Date: 6/5/2024
  * Author: Andreas Stergiopoulos (andreas.stergiopoulos@tuni.fi)
  *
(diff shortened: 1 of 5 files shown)


examples/headsail-bsp/src/apb_timer.rs
```diff
@@ -1,56 +1,77 @@
...
-#[inline]
-pub fn timer0_enable() {
-    // Read register
-    let mut reg = read_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET);
-    // Make enable bit 1
-    reg.set_bit(TIMER0_ENABLE_BIT, true);
-    // Write register back
-    write_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET, reg);
-}
+pub struct APBTimer<const BASE_ADDRESS: usize>;
```
hegza: Struct naming convention is PascalCase. Abbreviations work like words as usual, e.g. `ApbTimer`.

examples/headsail-bsp/src/apb_timer.rs
```diff
@@ -1,56 +1,77 @@
...
-#[inline]
-pub fn timer0_disable() {
-    // Read register
-    let mut reg = read_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET);
-    // Write 0 to bit 0 but leave all other bits untouched
-    reg.set_bit(TIMER0_ENABLE_BIT, false);
-    // Write register back
-    write_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET, reg);
-}
+impl<const BASE_ADDRESS: usize> APBTimer<BASE_ADDRESS>{
```
hegza: Spacing is strange. Please run the code through rustfmt so we can avoid back-and-forth changes. Consider configuring your editor to run cargo fmt automatically on save.

examples/headsail-bsp/src/apb_timer.rs
```diff
@@ -1,56 +1,77 @@
...
+     * Enables the timer (starts counting).
+     */
+    #[inline]
+    pub fn enable() {
+        // Read register
+        let mut reg = read_u32(BASE_ADDRESS + Self::TIMER_CTRL_REG_OFFSET);
+        // Make enable bit 1
+        reg.set_bit(Self::TIMER_ENABLE_BIT, true);
+        // Write register back
+        write_u32(BASE_ADDRESS + Self::TIMER_CTRL_REG_OFFSET, reg);
```
hegza: Commentary is a bit verbose all around. I'd leave out at least the "read register" and "write register" parts as they're already more than obvious from the source code.

examples/headsail-bsp/src/apb_timer.rs
```diff
@@ -1,56 +1,77 @@
...
-    return read_u32(TIMER0_ADDR + TIMER0_CTRL_REG_OFFSET);
+    /**
+     * Disables the timer (stops counting).
+     */
+    #[inline]
+    pub fn disable() {
+        // Read register
+        let mut reg = read_u32(BASE_ADDRESS + Self::TIMER_CTRL_REG_OFFSET);
+        // Write 0 to bit 0 but leave all other bits untouched
+        reg.set_bit(Self::TIMER_ENABLE_BIT as usize, false);
```
hegza: Superfluous cast. `TIMER_ENABLE_BIT` is already a usize.

examples/headsail-bsp/src/lib.rs
```diff
@@ -4,6 +4,16 @@
...
+    /*!
+     * Timer module for Headsail. When running on the Renode 
+     * Virtual Prototype, the "vp" feature should be enabled. 
+     * All operations are Read-Modify-Write.
+     * 
+     * HOW TO USE THIS DRIVER:
+     * In order to use any of the four timers that come with 
+     * Headsail HPC SubSystem, use the respective Timer{0..3} 
+     * type alias provided and the functions associated with it.
+     */
```
hegza, line 16: Very nice.

hegza: Please verify using `cargo doc --open` that this looks like how you intend it to look like. This outer attribute applies to the `timer` module and might compete with the outer attribute defined inside the module file.

examples/headsail-bsp/src/mmap.rs
```diff
@@ -1,2 +1,5 @@
 pub(crate) const UART0_ADDR: usize = 0xFFF00000;
-pub(crate) const TIMER0_ADDR: usize = 0x5_0000;
\ No newline at end of file
+pub(crate) const TIMER0_ADDR: usize = 0x5_0000;
+pub(crate) const TIMER1_ADDR: usize = 0x5_0010;
+pub(crate) const TIMER2_ADDR: usize = 0x5_0020;
+pub(crate) const TIMER3_ADDR: usize = 0x5_0030;
```
hegza, line 5: Are these memory maps correct for both VP and ASIC?

hegza (changes requested):
Looks good, only minor issues.

examples/headsail-bsp/src/mmap.rs
andstepan: Yes, those are the addresses provided in the GitLab pages relevant documentation.

examples/headsail-bsp/src/lib.rs
andstepan: Have already tested it, this is the comment that is shown in the documentation pages. As far as the outer comments in the respective driver files, I am a bit lost on why they are not showing in the documentation, as I have already marked them as outer comments. I think I need your help in that as I am probably missing something here.

examples/headsail-bsp/src/apb_timer.rs
andstepan: Done!

examples/headsail-bsp/src/lib.rs
hegza: I made a mistake with outer/inner; they're the other way around. I'll use the right terminology going forward.

I think this inner attribute here applies to the `mod timer` and supersedes the inner attribute specified inside the module (which also applies to `mod timer`) and that might be why you're not seeing it.

examples/headsail-bsp/src/apb_timer.rs
andstepan: Done!

examples/headsail-bsp/src/apb_timer.rs
andstepan: Done!

examples/headsail-bsp/src/apb_timer.rs
andstepan: Done!

examples/headsail-bsp/src/lib.rs
andstepan: Then in that case, by removing the comment in the lib.rs file would result in the comment inside the timer_unit.rs to show, but this is not the case I see when I run `cargo doc --open -Fvp`.

hegza: Alright, I'll investigate this to be able to give you a full answer. I won't hold up this merge though, so feel free to merge as the rest of the issues are done. Let's talk about this next week!

hegza (approved):

examples/headsail-bsp/src/lib.rs
hegza: I tried the documentation generation. Apparently rustdoc always takes the module documentation from the first definition at `lib.rs` even if it's empty. Therefore the in-file module-level documentation never makes it to the docs, as it's overridden by the re-export trickery.

EDIT: ah, I see. rustdoc is behaving exactly as instructed. `mod timer` does not "re-export the whole module" but instead just all definitions inside it. I'll see if I can think this through.

PR #39, opened 2024-05-31
Feat/clippy 2

vilukissa68:
Adds clippy style checks for headsail bsp and dla. This should help make PR reviews less focused on simple style issues. To make sure PR passes tests I ran `cargo clippy --fix` to automatically fix style issues in bsp and dla.


hegza (approved):
Looks great. We might want to run the linter with `-Fhpc-rt` to assure all HPC related code gets checked as well. But running it without preference for any feature is already good enough.

PR #40, opened 2024-06-04
Adds validation test for dla conv2d

vilukissa68:
Sorry for the large PR. Though, I feel that this time most of it is justified :D

PR adds `validate` example to dla-driver which implements one of the RTL conv2d tests. During this implementation, it was revealed how DLA maps memory to convolution operands wasn't as straightforward as it initially seemed. This required changing the DLA VP implementation, which is now validated against the RTL test.

To include RTL test data into the binary I wrote a script that converts .mem files from the RTL project to Rust arrays, which could then be included in the example.

Also adds robot script to run validation in CI.


.github/workflows/validate.yml
```diff
@@ -101,6 +102,48 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Download artifact
+      uses: actions/download-artifact@v4
+      with:
+        name: $DLA_VALIDATION_BIN
+    - name: Create Renode peripheral symlinks
+      run: ln -s $(readlink -f "./vp/devel/python_peripherals/DLA.py") "$RENODE_DIR/scripts/pydev/DLA.py"
+    - name: Run dla validation
+      run: renode-test scripts/robot/dla_validate.robot --variable BIN:"$(readlink -f $DLA_VALIDATION_BIN)"
```
hegza, line 140: This job creates a considerable amount of log output which is a probable strain on limited cloud resources over the long term. Browsing through it, it looks like some of it may not be that useful for someone looking to debug the job over CI.

Therefore, I would suggest one or more of the following:

1. Consider what prints are actually useful in general, and eliminate ones that are not
2. If a log is useful locally but rather useless in CI runs, hide it behind an environment variable, e.g. "CI=1". In Rust code: `if env!("CI").unwrap_or_default(0) == 1 { sprintln!("log text"); }`

There is also a way to centralize log orchestration via a crate called `log` but I think you'd prefer to just get along with the task, and there the above options are easier to set up.

examples/hpc/dla-driver/examples/test_data/conv_16x16x16_3x3_din.rs
hegza, line 1: FYI these could also be in a file such as "conv_16x16x16_3x3.in.data", which could be loaded in to a variable at compile time: `let in_data = include_bytes!("conv_16x16x16_3x3.in.data");`

What you did works too.

scripts/robot/dla_validate.robot
hegza, line 1: Consider using the commonly defined pass tag: `[PASS]` defined in headsail-bsp/src/tb.rs. This will make the test compatible with other CI scripts used in SoC Hub.

examples/hpc/dla-driver/examples/validate_conv.rs
```diff
@@ -0,0 +1,193 @@
...
+
+fn validate_conv2d_tiny() -> bool {
+    let mut dla = Dla::new();
+
+    let mut din: Vec<i8> = [
+        0, 0, 0, 2, 0, 0, 1, 2, 1, 2, 0, 0, 1, 2, 0, 1, 0, 0, 0, 2, 0, 0, 1, 0, 1, 2, 0, 1, 0, 1,
+        0, 0, 2, 2, 1, 1, 0, 2, 1, 1, 2, 1, 2, 2, 1, 0, 0, 1, 1, 2, 0, 1, 1, 1, 0, 0, 2, 0, 1, 2,
+        1, 0, 0, 1, 2, 1, 1, 1, 0, 0, 1, 1, 2, 0, 2,
+    ]
+    .to_vec();
```
hegza: Typically written `vec![0, 0, ...]`. I wonder if the vec macro's not available in this context?

examples/hpc/dla-driver/examples/validate_conv.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    let mut dout: Vec<i32> = [
+        -10, -1, -10, 0, -14, 2, -14, -4, -6, -5, -13, 4, -12, -2, -7, 1, -10, 0,
+    ]
+    .to_vec();
+
+    // Calculate output size
+    let (output_width, output_height) =
+        calculate_conv2d_out_param_dim((5, 5), (3, 3), (0, 0), (1, 1), (1, 1));
```
hegza: I'm a little bit surprised that this function does not already return `usize` as you keep doing the manual cast.

examples/hpc/dla-driver/examples/validate_conv.rs
```diff
@@ -0,0 +1,193 @@
...
+    } else {
+        sprintln!("16x16x16_3x3 conv2d test failed")
+    }
+
+    if succesful_test == 2 {
+        sprintln!("All tests succesful!\r\n");
+    } else {
+        sprintln!("Not all tests succesful!");
+    }
+
```
hegza: Consider using standard tags `[FAIL]` for "any failure" and `[PASS]` for "all test cases passed", perhaps `[OK]` for a single test ok.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -237,15 +240,24 @@ impl Dla {
...
+         */
+        for (cidx, chunk) in buf.chunks(8).enumerate() {
+            for (i, b) in chunk.iter().rev().enumerate() {
+                unsafe {
+                    ptr::write_volatile(
+                        (MEMORY_BANK_BASE_ADDR + offset + cidx * 8 + i) as *mut _,
+                        *b,
+                    )
+                };
+            }
```
hegza, line 255: Good comment, thank you.

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -285,28 +297,82 @@ impl Dla {
...
+    }
+
+    /// Reads len amount of bytes from DLA's output bank(s)
+    pub fn read_output_i8(&self, len: usize) -> Vec<i8> {
+        let bytes = self.read_data_bank(self.get_output_bank(), len);
+        bytes.iter().map(|&x| x as i8).collect()
+    }
+
+    /// Reads len amount of bytes from DLA's output bank(s)
+    pub fn read_output_i4(&self, len: usize) -> Vec<i8> {
```
hegza, line 333: Curious method. I guess this is used by something?

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -285,28 +297,82 @@ impl Dla {
...
         }
-        self.read_data_bank(MemoryBank::Bank0, len)
+        result
     }
 
     /// Reads len amount of bytes from DLA's input bank(s)
     pub fn read_input_bank(&self, len: usize) -> Vec<i8> {
-        self.read_data_bank(self.get_input_bank(), len)
+        let bytes = self.read_data_bank(self.get_input_bank(), len);
+        bytes.iter().map(|&x| x as i8).collect()
```
hegza, line 362: `.map(i8::from)` is arguably more readable.

scripts/convert-memt-to-rust.py
hegza, line 1: There must be an easier way :D but I'm willing to go with this if it's tested working.

hegza (changes requested):
Minor changes.

.github/workflows/validate.yml
vilukissa68: The terminal prints were caused by the DLA.py. These are generally useful for debugging the DLA behavior. I added a check to DLA which suppresses all Python prints if `CI` environment variable is set.

examples/hpc/dla-driver/examples/test_data/conv_16x16x16_3x3_din.rs
vilukissa68: Not implemented for now

scripts/robot/dla_validate.robot
vilukissa68: `[PASS]` tag added

examples/hpc/dla-driver/examples/validate_conv.rs
vilukissa68: The macro was not available by default, but adding
```rust
#[macro_use]
extern crate alloc;
```
enabled use of the `vec!` macro. This style has been adopted to validation test.

examples/hpc/dla-driver/examples/validate_conv.rs
vilukissa68: `calculate_conv2d_out_param_dim` now returns `(usize, usize)`

examples/hpc/dla-driver/examples/validate_conv.rs
vilukissa68: `[FAIL]` and `[PASS]` tags are now used as suggested. Extended `tb.rs` file with `report_ok` method to allowing using `[OK]`tag.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: No problem! :D

examples/hpc/dla-driver/src/lib.rs
vilukissa68: This is used to read and unpack 4-bit SIMD calculations from DLA. In short it's possible to do calculation on DLA by packing two or four values to each byte by reducing the bit width of input and kernel values. This allows for more parallelism with the cost of granularity in results. My idea here was to unpack each byte to two i8's since as per my understanding i4 and i2 types aren't supported.

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Apparently `i8`from `u8`is not supported due to numerical value changing, so I implemented this with `try_from::i8`.
```rust
bytes.iter().map(|&x| i8::try_from(x).unwrap()).collect()
```

scripts/convert-memt-to-rust.py
vilukissa68: There probably is, but I don't think it's worth spending anymore time to figure it out. The problem comes from the fact that RTL code can easily interpret the hex values as signed, where as Rust and Python only allow conversion to unsigned.

examples/headsail-bsp/src/tb.rs
```diff
@@ -4,6 +4,7 @@ use crate::uart::uart_write;
 
 pub const TAG_FAIL: &str = "[FAIL]";
 pub const TAG_PASS: &str = "[PASS]";
+pub const TAG_OK: &str = "[OK]";
```
hegza, line 9: Please document this with a `///`. Make sure the difference between [OK] and [PASS] is clear.

examples/hpc/dla-driver/src/lib.rs
hegza: Oh, right. That's true. Anyway, the try_from().unwrap() does not really belong to the driver. I would much rather return exactly whatever comes from the register or possibly do an infallible conversion.

If the hardware defines how out-of-range values are interpreted, interpret them like that. If the hardware promises to only return values in valid i8 range, use `as` as before. The `as` operator in this case does an infallible 2's complement conversion sa. https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast.

Sorry about the back-and-forth.

hegza (approved):
Need to resolve these but merge afterwards. I'll set it this approve as I don't think a 3rd review will be of additional utility.

PR #42, opened 2024-06-14
Adds ability to read from uart to heap

vilukissa68:
In preparation to DLA inputs being sent to HPC over UART, we need to have ability read UART in headsail-bsp. PR adds basic functionality for reading known number of bytes from UART into heap.

### DIFF
--- a/examples/headsail-bsp/src/mmap.rs
+++ b/examples/headsail-bsp/src/mmap.rs
@@ -1,6 +1,3 @@
-#[cfg(feature = "hpc")]
-pub(crate) const UART0_ADDR: usize = 0xFFF00000;
-#[cfg(not(feature = "hpc"))]
 pub(crate) const UART0_ADDR: usize = 0xFFF00000;
 // NOTE: (20240614 vaino-waltteri.granat@tuni.fi) This applies to renode NS16550 uart, but might not apply to headsail ASIC
 pub(crate) const UART_DATA_READY_OFFSET: usize = 5;
--- a/examples/headsail-bsp/src/mmap.rs
+++ b/examples/headsail-bsp/src/mmap.rs
@@ -1,4 +1,9 @@
+#[cfg(feature = "hpc")]
 pub(crate) const UART0_ADDR: usize = 0xFFF00000;
+#[cfg(not(feature = "hpc"))]
+pub(crate) const UART0_ADDR: usize = 0xFFF00000;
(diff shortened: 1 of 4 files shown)


examples/headsail-bsp/src/mmap.rs
```diff
@@ -1,4 +1,6 @@
 pub(crate) const UART0_ADDR: usize = 0xFFF00000;
+// NOTE: (20240614 vaino-waltteri.granat@tuni.fi) This applies to renode NS16550 uart, but might not apply to headsail ASIC
```
hegza, line 2: Consider annotating with `#[cfg(feature = "vp")]`.

examples/headsail-bsp/src/uart.rs
```diff
@@ -12,3 +17,26 @@ pub fn putc(c: u8) {
...
+
+#[inline]
+pub fn getc() -> Option<u8> {
+    let data_ready = unsafe { read_u8(UART0_ADDR + UART_DATA_READY_OFFSET) };
+    if data_ready & 1 == 0 {
+        None
+    } else {
+        let byte = unsafe { read_u8(UART0_ADDR) };
+        Some(byte)
+    }
```
hegza: Minor, I recommend writing this like so:

```rust
// Data is not yet ready
if data_ready & 1 == 0 {
    return None;
}

Some(unsafe { read_u8(UART0_ADDR) });
```

De-nested code tends to be easier to read for most people.

hegza: I wonder if this should be more along the lines of 

```rust
// Wait for data to become ready
while data_ready & 1 == 0 {}

// SAFETY: UART0_ADDR is 4-byte aligned
unsafe {read_u8(UART0_ADDR)};
```

This will also simplify caller at `uart_read_to_heap`.

hegza (approved):
Consider the feedback if relevant. It's all good like this though.

hegza:
Commendation for the excellent PR scope.

examples/headsail-bsp/src/uart.rs
vilukissa68: Added the latter suggestion since it looks very clean

examples/headsail-bsp/src/mmap.rs
vilukissa68: Didn't implement this, since it's still unclear if this is VP only

PR #41, opened 2024-06-11
Feat/external bit

vilukissa68:
Adds external bit addressing for HPC in VP and BSP.  This solution modifies read and write methods in BSP so that all HPC operation addresses are preceded with `0x1_0000_0000` to ensure similar functionality as described in issue https://github.com/soc-hub-fi/headsail-vp/issues/33. New addresses in drivers should be defined from the Sysctrl point of view, which then gets automatically addressed with the external bit on HPC. These changes required the DLA driver to utilize BSP read/write operation, which was not the case before.

Previously VP was described in `common.repl`, `hpc.repl` and `sysctrl.repl`, due to new dependencies imposed by the shadowed sdram these files have been concatenated into one `headsail.repl` file, which now encompasses the whole VP. HPC memory regions are now preceded by `0x1_0000_0000`.

PR #44, opened 2024-06-17
Adds support for HPC external bit in DLA

vilukissa68:
In https://github.com/soc-hub-fi/headsail-vp/pull/41 it was agreed to leave DLA broken to retain a somewhat manageable scope for the PR. This PR adds HPC external bit handling to the DLA.py and dla-driver. The conv2d validation is not fixed in this PR.


hegza (approved):
Seems reasonable. Review borderline necessary :)

PR #45, opened 2024-06-19
Moves to use riscv64 linker

vilukissa68:
`riscv64-unknown-elf-ld` allows us to link a binary to addresses > 0xFFFF_FFFF, so we should switch to using that. This exposed an error in `sdram_hpc.x` where the memory started too early. This has been fixed to reflect SDRAM defined in `heasail.repl`.


examples/headsail-bsp/sdram_hpc.x
```diff
@@ -1,12 +1,12 @@
 MEMORY
 {
   BOOTRAM : ORIGIN = 0x10000, LENGTH = 0x8000
-  SDRAM : ORIGIN = 0x100000000, LENGTH = 0x70000000
+  SDRAM : ORIGIN = 0x104080000, LENGTH = 0x1bf80000
```
hegza: We think the SDRAM is actually available starting at 0x1_0000_0000. The memory map just has some incorrect tags that overlap it. Those tags need to be corrected or disabled.

examples/headsail-bsp/sdram_hpc.x
```diff
@@ -1,12 +1,12 @@
 MEMORY
 {
   BOOTRAM : ORIGIN = 0x10000, LENGTH = 0x8000
```
hegza: This `BOOTRAM` line should be removed if it's not used in the rest of the script.

hegza (changes requested):
There's still some trouble with the memory map.

examples/headsail-bsp/sdram_hpc.x
vilukissa68: `BOOTRAM` removed

examples/headsail-bsp/sdram_hpc.x
vilukissa68: Memory map adjusted as discussed + HPC bootram external addressing from HPC fixed in headsail.repl

PR #46, opened 2024-06-20
Feat/dla uart

vilukissa68:
DLA validation test has been broken due to the memory limitations discussed previously. This PR modifies DLA validation to use UART instead of binary embedded inputs, weights, and outputs, which fixes the problem with CI. To do this in CI I added a UartLibrary.py to Robot framework which reads the new .mem files and writes them over UART to headsail. This approach should streamline the process of adding new DLA tests to the project since data can be sent almost arbitrarily. The size of the PR might look big but most of is comes from the new inputs and weights.

PR #48, opened 2024-07-09
Removes unused dla validation data files

vilukissa68:
#46 removed the need to embed DLA testing data directly to the binary. Removes these files as they are no longer needed.


hegza (approved):
All good, merge whenever. Having the CI in passing state is still good policy to enable tracking regression.

PR #49, opened 2024-07-26
Feat/padding fix

vilukissa68:
Fixes convolution to support padding in all 4 directions. Previously padding was expected to be uniform per axis. Locally run validation test still passes.


hegza (approved):
Yeah, seems good to me. Recommend passing CI however.

PR #50, opened 2024-07-26
Feat/bias

vilukissa68:
Adds bias support for dla-driver and dla-vp with a validation test.

This should be reviewed and merged after #46, #48 and #49 have been merged.

PR #51, opened 2024-07-29
Feat/dla high level

vilukissa68:
Adds higher level API to dla to provide direct operation calls for conv2d, relu, bias and fusion layers. This is required for TVM codegen integration. #50 and dependant PRs should be merged before this.


hegza (changes requested):
Please resolve conflicts & rebase.

vilukissa68:
Split Tensors out of this PR to #54, so that should be review first.

vilukissa68:
Split layer API to #55, so that should be review before this

vilukissa68:
After extracting Tensors and the Layer API out, this PR was left with the high-level example and validation test that uses the new API with Tensors.

examples/hpc/dla-driver/examples/highlevel.rs
```diff
@@ -0,0 +1,70 @@
...
+use headsail_bsp::{init_alloc, rt::entry, sprint, sprintln};
+use panic_halt as _;
+
+use dla_driver::tensor3::{Order3, Tensor3};
+use dla_driver::tensor4::{Order4, Tensor4};
+
+use alloc::vec::Vec;
+
+fn dense_test() {
+    sprintln!("Starting dense test");
```
hegza: If you'd like some consistency with the start and end prints, here's what I use:

> dense_test: enter
> dense_test: leave
> conv_test: enter
> conv_test: leave

examples/hpc/dla-driver/src/utils.rs
hegza, line 1: Please document the functions in this file. _util_ functions like these are typically low in context and their purpose is hard to interpret without appropriate documentation.

- https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments
- https://stackoverflow.com/a/30013052

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -804,6 +820,7 @@ impl Dla {
             self.handshake_disable_hw();
             return false;
         }
+        sprintln!("Finishing handshake");
```
hegza: Consider if prints in driver code are necessary or if they should be put behind a feature flag :)

examples/hpc/dla-driver/src/lib.rs
```diff
@@ -16,9 +21,9 @@ pub use mmap::{
 };
 
 const DEFAULT_INPUT_BANK: MemoryBank = MemoryBank::Bank0;
-const DEFAULT_KERNEL_BANK: MemoryBank = MemoryBank::Bank8;
-const DEFAULT_OUTPUT_BANK: MemoryBank = MemoryBank::Bank12;
-const DEFAULT_BIAS_ADDR: u32 = 0x0;
+const DEFAULT_KERNEL_BANK: MemoryBank = MemoryBank::Bank4;
+const DEFAULT_OUTPUT_BANK: MemoryBank = MemoryBank::Bank10;
+const DEFAULT_BIAS_ADDR: u32 = (MEMORY_BANK_BASE_ADDR + MEMORY_BANK_15_OFFSET) as u32;
```
hegza: This looks like it should be a bank as well. Isn't `DEFAULT_BIAS_ADDR` effectively set to `Bank15` now? In case you need to retrieve the address from a `MemoryBank` instance, consider adding/using a method like `addr()` on `MemoryBank`.

examples/hpc/dla-driver/examples/validate_conv.rs
hegza, line 1: The new API looks reasonable and is more readable to me at least. One thing that the high-level API seems to lose is the bank selection. Is it important to be able to customize the bank selections and is it possible to do with the new API?

hegza (changes requested):
The new API feels nicer to use and the increased lucidity is visible in the examples. Utility functions need a little bit of documentation.

examples/hpc/dla-driver/examples/validate_conv.rs
vilukissa68: As long as all the data fits into the banks I don't see why the user might want to select banks themselves. There is an optimization that can be done when the previous layer's output is directly used as the following layer's input, where swapping the output bank and input bank, can eliminate the need for writing the input data. Ideally, this is something that the high-level interface handles automatically. This is not done yet, since I'm not sure if it's possible to determine from IR graph the data dependency.

examples/hpc/dla-driver/examples/highlevel.rs
vilukissa68: Fixed

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Print removed

examples/hpc/dla-driver/src/lib.rs
vilukissa68: Modified to use `addr()` to retrieve `MemoryBank` address for setting default bias

examples/hpc/dla-driver/src/utils.rs
vilukissa68: Added documentation for util functions

examples/hpc/dla-driver/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+/// * `padding` - Padding used in the given layer.
+/// * `stride` - Stride used in the given layer.
+pub fn generate_output_tensor<T: Clone, H: Clone, J: Clone>(
+    input: &Tensor3<T>,
+    kernel: &Tensor4<H>,
+    output_buf: Vec<J>,
+    order: Order3,
+    padding: Option<Padding>,
+    stride: Option<Stride>,
+) -> Tensor3<J> {
```
hegza: Generic names `I, K, O` might be more memorizable.

examples/hpc/dla-driver/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+/// * `bias_size` - The size of bias in bytes. If left empty bias is not assigned.
+pub fn get_banks_for_layer(
+    input_size: usize,
+    kernels_size: usize,
+    output_size: usize,
+    bias_size: Option<usize>,
+) -> (MemoryBank, MemoryBank, MemoryBank, Option<u32>) {
+    let no_input_banks = calculate_number_of_banks_needed(input_size);
+    let no_kernel_banks = calculate_number_of_banks_needed(kernels_size);
+    let no_output_banks = calculate_number_of_banks_needed(output_size);
```
hegza: `nr` or `num` better than `no`.

examples/hpc/dla-driver/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let input_bank = MemoryBank::Bank0;
+    let kernel_bank = input_bank + no_input_banks;
+    let output_bank = kernel_bank + no_kernel_banks;
+
+    let bias_bank = bias_size
+        .map(|_| (MEMORY_BANK_BASE_ADDR + (output_bank + no_output_banks).offset()) as u32);
```
hegza: `bias_size` unused in this expression. Look like the following would work here:

```
let bias_bank = MEMORY_BANK_BASE_ADDR + (output_bank + no_output_banks).offset()) as u32;
```

examples/hpc/dla-driver/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+    (input_bank, kernel_bank, output_bank, bias_bank)
+}
+
+/// Divides x with y and ceils the output
+fn ceil<T>(x: T, y: T) -> T
```
hegza: `div` or `ceil_div` feels more conventional here as "div" is the performed operation.

examples/hpc/dla-driver/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+        + Copy,
+{
+    (x + y - T::from(1)) / y
+}
+
+/// Calculates the padding needed to produce output with the same size as the input
+/// * `input` - Input data for a given layer.
+/// * `kernel` - Kernels/weight data for a given layer.
+/// * `stride` - Stride used in the given layer.
+fn calculate_same_padding(input: (u32, u32), kernel: (u32, u32), stride: Stride) -> Padding {
```
hegza, line 112: `calculate_required_padding`? The word "same" seems weird here :)

hegza (approved):
Util functions are now well documented and I've understood the purpose of each. Minor naming things only.

Correct if you feel like it, then press merge.

examples/hpc/dla-driver/src/utils.rs
vilukissa68: In Tensorflow `same` signifies the amount of padding that keeps input and output layers the same size, hence why `same` is used here.

hegza: Ok, seems reasonable then.

examples/hpc/dla-driver/src/utils.rs
vilukissa68: Fixed

examples/hpc/dla-driver/src/utils.rs
vilukissa68: Fixed

examples/hpc/dla-driver/src/utils.rs
vilukissa68: fixed

examples/hpc/dla-driver/src/utils.rs
vilukissa68: The suggestion doesn't seem to work due to `bias_bank` being wrapped into `option`

Issue #35, opened 2024-05-22
Validate memory map against results from RTL sim


hegza:
We've got results from the memory map validation on the RTL sim using `keelhaul`. The results should match exactly on the VP, so let's run [keelhaul](https://github.com/soc-hub-fi/keelhaul/)  against the VP and correct any differences.

Latest tested SVDs for Headsail are found at: https://gitlab.tuni.fi/soc-hub/headsail/hw/headsail/-/tree/main/verification/common/svd. We want to run this for HPC and SysCtrl.

hegza:
Keelhaul memory map test utility added at <https://github.com/soc-hub-fi/headsail-vp/tree/main/examples/hpc/test-memory-maps>. Use `just run <device>` to see in Renode, what registers don't match what was tested on RTL, e.g. `just run apb_uart0`.

PR #53, opened 2024-07-30
CI: memory map tests

hegza:


hegza:
Fixes #35

hegza:
We fail to catch the failures currently. We'd need to capture and match output from Renode to detect invalid accesses by software. An attempt was made in this PR using Robot but it doesn't seem to work,

PR #54, opened 2024-07-31
Adds Tensors objects to DLA

vilukissa68:
#51 Turned out to be way too big to review, so I split tensor functionality to this PR to reduce the amount of code to review.

Tensors are used to abstract the shape of data. This is useful since DLA expected a different data shape than what TVM does. Also since different models might use other data formats it is sensible to generalize tensors to all possible data shapes.

### DIFF
--- a/examples/hpc/dla-driver/src/tensor3.rs
+++ b/examples/hpc/dla-driver/src/tensor3.rs
@@ -15,7 +15,7 @@ pub enum Order3 {
 }
 
 impl Order3 {
-    fn into_position(&self) -> [usize; 3] {
+    fn into_position(self) -> [usize; 3] {
         match self {
             Order3::CHW => [0, 1, 2],
             Order3::CWH => [0, 2, 1],
--- a/examples/hpc/dla-driver/src/tensor3.rs
+++ b/examples/hpc/dla-driver/src/tensor3.rs
@@ -2,8 +2,6 @@
 #![no_main]
 
 use alloc::vec::*;
-use headsail_bsp::ufmt::uDisplay;
-use headsail_bsp::{sprint, sprintln};
 use ndarray::{Array, Array3};
 
 #[derive(Clone, Copy, Debug, PartialEq)]
@@ -17,8 +15,8 @@ pub enum Order3 {
 }
 
 impl Order3 {
-    fn into_positions(order: Order3) -> [usize; 3] {
-        match order {
+    fn into_position(&self) -> [usize; 3] {
+        match self {
             Order3::CHW => [0, 1, 2],
             Order3::CWH => [0, 2, 1],
             Order3::HWC => [1, 2, 0],
@@ -35,7 +33,7 @@ pub struct Tensor3<T> {
     order: Order3,
 }
 
-impl<T: Clone + uDisplay> Tensor3<T> {
+impl<T: Clone> Tensor3<T> {
     // Creates a new Tensor3 with the specified dimensions, initial value, and order
     pub fn new(
         channels: usize,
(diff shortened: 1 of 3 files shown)


examples/hpc/dla-driver/src/tensor3.rs
```diff
@@ -0,0 +1,248 @@
...
+        match order {
+            Order3::CHW => [0, 1, 2],
+            Order3::CWH => [0, 2, 1],
+            Order3::HWC => [1, 2, 0],
+            Order3::HCW => [1, 0, 2],
+            Order3::WHC => [2, 1, 0],
+            Order3::WCH => [2, 0, 1],
+        }
+    }
+}
```
hegza: `From`/`Into` is for 1:1 type conversions. If converting to a specific kind of representation it may be more clear to use something like `Order3::as_positions` / `into_positions`. This allows for multiple implementations, e.g. `fn into_tensorflow_order`, `fn into_reverse_tensorflow_order`, `fn into_custom_order`.

examples/hpc/dla-driver/src/tensor3.rs
```diff
@@ -0,0 +1,248 @@
...
+
+#[derive(Clone, Debug)]
+pub struct Tensor3<T> {
+    data: Array3<T>,
+    pub channels: usize,
+    pub height: usize,
+    pub width: usize,
```
hegza: `data` seems to already contain the information for `channels`, `height`, `width`. These three fields are now duplicated between the inner struct (`Array3`) and outer struct (`Tensor3`) and need to be kept in sync everywhere which isn't great.

hegza: In case you need easy access to the symbolic `channels`, `height`, `width` values (which seem `order`-dependent), replace these three fields with three functions of the same name, which are able to resolve the correct dimension based on the `order` field.

examples/hpc/dla-driver/src/tensor3.rs
```diff
@@ -0,0 +1,248 @@
...
+            return self.to_buffer();
+        }
+
+        let mut data = self.clone();
+        data.transmute(order);
+        data.to_buffer()
+    }
+
+    /// Prints tensor in current order
+    pub fn print_tensor(&self) {
```
hegza: This is fine for debug but on bare-metal, sprintln is typically very runtime dependent. I would generally avoid depending on sprintln directly from any driver components to enable easy porting, i.e., avoiding any `print_*` methods. This type of design is by the way exactly the reason why TVM runtime requires newlib; it's a design choice. Maybe more along the lines of _"something to keep in mind"_ than _"needs fix"_.

Fixable in many ways, all optional:

1. An easy alternative is to extract this method out of the struct.
2. Putting it behind a feature flag also makes this (arguably) easier to port.
3. One alternative would be to add `T: core::fmt::Write` as parameter then write into that instead of using sprintln. Then this could be used like so `tensor.write(&mut serial);`

examples/hpc/dla-driver/src/tensor4.rs
hegza, line 1: All feedback that applies to tensor3 applies to tensor4 as well.

examples/hpc/dla-driver/src/tensor4.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+    }
+
+    /// Returns the dimensions of the array
+    pub fn dimensions(&self) -> (usize, usize, usize, usize) {
+        (self.kernels, self.channels, self.height, self.width)
+    }
+
+    /// Sets a new order for the array
+    pub fn transmute(&mut self, order: Order4) {
```
hegza: `permute`?

hegza (changes requested):
Some ideas for improvements to avoid future stumbles.

examples/hpc/dla-driver/src/tensor3.rs
vilukissa68: Moved this functionality to `into_position` function

examples/hpc/dla-driver/src/tensor3.rs
vilukissa68: Added functions to access channels, height, and width.

examples/hpc/dla-driver/src/tensor3.rs
vilukissa68: Yes, this is definitely a good guideline to follow. Removed all printing from tensors.

examples/hpc/dla-driver/src/tensor4.rs
vilukissa68: Made changes for tensor3 and tensor4

examples/hpc/dla-driver/src/tensor4.rs
vilukissa68: Renamed to `permute`

vilukissa68:
Tested suggested changes with #51 and everything seems to work

examples/hpc/dla-driver/src/tensor3.rs
```diff
@@ -0,0 +1,207 @@
...
+        initial_value: T,
+        order: Order3,
+    ) -> Self {
+        let data = Array::from_elem((channels, height, width), initial_value);
+        Tensor3 { data, order }
+    }
+
+    pub fn channels(&self) -> usize {
+        let dim_order: [usize; 3] = self.order.into_position();
+        let position = dim_order.iter().position(|&r| r == 0).unwrap();
```
hegza: I would discourage use of `.unwrap()` in drivers which can pollute the generated code with panic implementations. They say a compiler is smart and that it would eliminate such things but based on experience, this is often **not** the case, and it's bitten me back on memory constrained implementations :)

If you know **for sure** that this unwrap will **never** fail, you can use `unwrap_unchecked()`. I think that is appropriate in this case since we know how these arrays are constructed and can guarantee that the values will always be `0`, `1` and `2`. Same for `height` and `width` functions.

If you think it can fail, `unwrap` is reasonable, though forwarding the problem by returning `Option<usize>` or `Error<usize, _>` is more appropriate.

hegza (approved):
Ideally, replace `unwrap` with `unwrap_unchecked` where fitting but please merge afterwards.

examples/hpc/dla-driver/src/tensor3.rs
vilukissa68: Okay, this makes sense. In this case there isn't a chance of failure, so I replaced `unwrap()` with `unwrap_unchecked()`.

PR #55, opened 2024-07-31
Adds API for DLA operations

vilukissa68:
Split from #51 to reduce the amount of reviewable code.

Implements a high-level API for calling DLA operations. This is necessary for TVM code generation. There might be a reason to do this with way less code, but I decided to go with a verbose solution i.e. one function per operation or fusion operation, since all the TVM code generation examples use this approach.

### DIFF
--- a/examples/hpc/dla-driver/src/layers.rs
+++ b/examples/hpc/dla-driver/src/layers.rs
@@ -214,10 +214,7 @@ fn run_layers<T: DlaOutputType + Clone>(
 
     let dla = Dla::new();
 
-    let bias_size = match bias {
-        Some(ref bias) => Some(bias.len()),
-        None => None,
-    };
+    let bias_size = bias.as_ref().map(|bias| bias.len());
 
     let banks = get_banks_for_layer(
         input.get_size(),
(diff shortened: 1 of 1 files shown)


hegza:
Took a cursory look I'll review this a bit later, just a quick preliminary comment.

- I agree that matching the DLA examples is likely a good choice; verbose is good for starters.
- The code is heavily duplicated, however. Consider if there exists a sensible, clear private `fn` to wrap any of the boilerplate. Current approach is fine too. It's easier to dedup than to refactor.
- If this API is very much TVM specific, consider placing the `pub fn`'s into a module such as `ops::tvm` or `tvm_ops`.

vilukissa68:
I tried to reduce duplication with a new private function which operation calls wrap.

examples/hpc/dla-driver/src/layers.rs
```diff
@@ -0,0 +1,277 @@
...
+use crate::utils::{
+    calculate_conv2d_out_param_dim, calculate_number_of_banks_needed, get_banks_for_layer,
+};
+
+#[derive(Clone, Copy)]
+enum DlaOutput {
+    I8,
+    I16,
+    I32,
+}
```
hegza: It looks like this is not used, at least in this file. I would remove this, and rename `DlaOutputType` as `DlaOutput`.

hegza (approved):
Minor unused symbol. Other than that, I think it's good for merge.

PR #56, opened 2024-08-02
Fix minor issues in DLA driver

hegza:
Started by looking at the option thing discussed in a recent PR, ended up fixing a bunch of warnings.

### DIFF
--- a/examples/hpc/dla-driver/src/utils.rs
+++ b/examples/hpc/dla-driver/src/utils.rs
@@ -1,5 +1,3 @@
-#![no_std]
-
 use crate::mmap::MEMORY_BANK_SIZE;
 use crate::tensor3::{Order3, Tensor3};
 use crate::tensor4::Tensor4;
--- a/examples/hpc/dla-driver/src/utils.rs
+++ b/examples/hpc/dla-driver/src/utils.rs
@@ -76,7 +76,6 @@ pub fn get_banks_for_layer(
     input_size: usize,
     kernels_size: usize,
     output_size: usize,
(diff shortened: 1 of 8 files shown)


hegza:
Check especially this commit which has the `Option` thing I mentioned in a recent PR: https://github.com/soc-hub-fi/headsail-vp/pull/56/commits/a84e9a4437369c73519f5cbd8cb8c1541abbc36b.

There was an extra parameter on `get_banks_for_layer` that wasn't used for anything, except for the side-effect that the type had a method capable of producing an option. You can also produce an option by just wrapping it in `Some(_)`.

hegza:
I noticed also that there was a little bit of a warning drown going on. It may be difficult to tell when there's a useful warning, when there's dozens of superfluous ones, therefore I took a stance and disabled dead_code for mmio.rs and fixed all the trivial ones.

I do generally recommend getting rid of dead_code altogether (e.g., unused functions). If you need them back later, use version control.

examples/hpc/dla-driver/src/utils.rs
```diff
@@ -78,7 +76,6 @@ pub fn get_banks_for_layer(
     input_size: usize,
     kernels_size: usize,
     output_size: usize,
-    bias_size: Option<usize>,
```
vilukissa68, line 81: Okay, now I understand what you meant in the previous PR. This makes way more sense.

vilukissa68 (approved):
Thanks for taking time to fix these! Moving on it probably makes more sense for me to fix this kind of stuff :D

PR #57, opened 2024-08-15
Feat/dla ffi

vilukissa68:
Adds FFI for DLA layer API to be used from C. This is necessary to run TVM acceleration since the TVM graph executor calls these bindings.

PR #58, opened 2024-08-15
Add test case for DLA FFI

vilukissa68:
Add example use case for DLA FII added in #57.


hegza (approved):
All good if it works in tests. I ran an auto-format for the code for no particular reason.

PR #61, opened 2024-08-20
Rename bsp::uart as bsp::apb_uart

hegza:
... to disambiguate from sysctrl::uart. Make it clear to any user that apb_uart0 and apb_uart1 are the top-level, shared UARTs.


hegza:
@vilukissa68 Can you think of anything that might break as a result of this? Feel free to merge if you think this is non-breaking.

vilukissa68 (approved):
Can't think of anything that would break with these changes. The C-examples mostly rely on Newlib for UART functionality, which this doesn't affect. If something broke on the rust side dla-validation would have failed.

PR #62, opened 2024-08-20
Relocatable UART

hegza:
As per heading. Added also safety valves for C-style code (`ApbUart0::instance`).

Also makes examples compatible with ASIC.


hegza:
Just wanted to make sure this passes by your eyes. Check `examples/hpc/dla-driver/examples/validate_conv.rs` and you should be able to gather the main idea. Then press squash & merge.

vilukissa68 (approved):
Looks good to me

PR #63, opened 2024-08-23
SysCtrl uDMA UART for ASIC

hegza:


hegza:
Memory maps in closed GitLab causing issues for CI. Might check if we can publish the memory maps to avoid having to jump these hurdles.

PR #69, opened 2024-09-24
Feat/dla vp overflow fix

vilukissa68:
Fixes buffer overflow issue with bank switching. Adds some quality of life features to prints. Removes dead code.

### DIFF
--- a/vp/devel/python_peripherals/DLA.py
+++ b/vp/devel/python_peripherals/DLA.py
@@ -837,20 +837,12 @@ class Dla:
                     for w in range(width):
                         idx = c + (f * input_channels) + (input_channels*filter_amount) * w + (input_channels * filter_amount * width) * h
                         column_wise.append(data[idx])
-        #print('[{}]'.format(', '.join("{:2}".format(hex(x & 0xff)[2:-1]) for x in column_wise)))
 
         column_wise = reshape(column_wise, (filter_amount, input_channels, height, width))
 
         print_matrix(column_wise[0][0], "flat kernel K0 C0:", pformat="decimal")
         print_matrix(column_wise[0][1], "flat kernel K0 C1:", pformat="decimal")
         print_matrix(column_wise[0][2], "flat kernel K0 C2:", pformat="decimal")
-        print_matrix(column_wise[1][0], "flat kernel K1 C0:", pformat="decimal")
-        print_matrix(column_wise[1][1], "flat kernel K1 C1:", pformat="decimal")
-        print_matrix(column_wise[1][2], "flat kernel K1 C2:", pformat="decimal")
-        print_matrix(column_wise[2][0], "flat kernel K2 C0:", pformat="decimal")
-        print_matrix(column_wise[2][1], "flat kernel K2 C1:", pformat="decimal")
-        print_matrix(column_wise[2][2], "flat kernel K2 C2:", pformat="decimal")
-
 
         return filter_amount, s_channels, width, height, column_wise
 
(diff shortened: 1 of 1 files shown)


vp/devel/python_peripherals/DLA.py
```diff
@@ -318,7 +295,7 @@ def cast_long_to_signed_byte(value):
     Returns:
     byte -- Int Signed value in range -128..127
     """
-    assert(0 <= value <= 255)
+    assert(0 <= value <= 255), "Assert failed! Value doesn't fit byte"
```
hegza: Might be better expressed as `assert(0 <= value <= 0xff)`

vp/devel/python_peripherals/DLA.py
```diff
@@ -333,7 +310,7 @@ def cast_long_to_signed_16(value):
     Returns:
     byte -- Int Signed value in range -128..127
     """
-    assert(0 <= value <= 65535)
+    assert(0 <= value <= 65535), "Assert failed! Value doesn't fit 2 bytes"
```
hegza: Same as above, `<= 0xffff`

vp/devel/python_peripherals/DLA.py
```diff
@@ -279,36 +280,12 @@ def construct(flat, shape):
...
-    output = [[[0 for _ in range(height)] for _ in range(width)] for _ in range(channels)]
-    i = 0
-    for ch in range(channels):
-        for w in range(width):
-            for h in range(height):
-                output[ch][w][h] = data[i]
-                i = i + 1
-    return output
-
 def cast_long_to_signed_byte(value):
```
hegza, line 289: Not part of the review, but I wonder, how did this C-language ABI-specific type names (long) end up in our source tree :sweat_smile: 

Ideally, this would be something like `cast_u64_to_i8`. Besides, current function name leaves it up to interpretation whether it's a signed or an unsigned long. One has to wonder, why is the `signed` name-specifier included for the byte but not for the long.

vp/devel/python_peripherals/DLA.py
```diff
@@ -1180,55 +1166,50 @@ def conv2d(self, input_img, kernels, padding, dilation=(1,1), stride=(1,1), padd
 
 
         # Apply each kernel to input_img
+        first_pass = True
```
hegza: unused?

vp/devel/python_peripherals/DLA.py
```diff
@@ -1274,7 +1255,7 @@ def matsum_element_wise(self, A, B):
         if not isinstance(A, list) and not isinstance(B, list):
             return A + B
 
-        assert len(A) == len(B)
+        assert len(A) == len(B), "Assert failed! Different sized operands in matsum"
```
hegza: Assert style changed with respect to earlier code. Seemed to have parentheses before but not here.

hegza (approved):
Seems reasonable. Mostly style issues. Fix if you will.

vp/devel/python_peripherals/DLA.py
vilukissa68: So the logic is that the input value is a Python long type, an integer of (almost) unlimited size. We want to sanitize the longs from Renode and reinterpret them to logical values in the range of -128..127. For naming, i8 would make sense here, but u64 is wrong. I don't know if this answers your question, but it's a necessary evil from working on a byte level in Python with Renode.

vp/devel/python_peripherals/DLA.py
vilukissa68: Fixed

vp/devel/python_peripherals/DLA.py
vilukissa68: Fixed

vp/devel/python_peripherals/DLA.py
vilukissa68: Removed

vp/devel/python_peripherals/DLA.py
vilukissa68: Style unified to have parentheses

PR #72, opened 2024-09-25
Add TVM compatible interface to dla-ffi

vilukissa68:
Our TVM codegen mandates a particular order for the external interface arguments and certain scaling operations to be done before and after qnn.conv2d call.

### DIFF
--- a/examples/hpc/dla-driver-ffi/src/lib.rs
+++ b/examples/hpc/dla-driver-ffi/src/lib.rs
@@ -275,11 +275,38 @@ pub unsafe extern "C" fn dla_conv2d_relu(
 }
 
 /// Executes Conv2D + Bias on DLA with given parameters and writes result to output buffer.
+/// # Arguments
+///
+/// * `input_data` - Input data  
+/// * `kernel_data` - Kernel/weight data
+/// * `bias` - Buffer containing bias data. NOTE: Bias is actually i16 in hardware, here we use 32 for TVM compatibility
+/// * `output` - Buffer for operation output
+/// * `input_channels` - Number of channels in the input
+/// * `input_height` - Height of the input
+/// * `input_width` - Width of the input
+/// * `input_order` - Layout of the input data
+/// * `input_zero` - Possible zero point correction value
+/// * `kernel_amount` - Number of output channels
+/// * `kernel_channels` - Number of input channels
+/// * `kernel_height` - Kernel height
+/// * `kernel_width` - Kernel width
+/// * `kernel_order` - Kernel data layout
+/// * `bias_length` - Size of the bias fifo
+/// * `pad_top` - Amount of padding on top of the data
+/// * `pad_right` - Amount of padding on right side of the data
+/// * `pad_left` - Amount of padding on left side of the data
+/// * `pad_bottom` - Amount of padding on bottom of the data
+/// * `pad_value` - Integer value used for padding the data
+/// * `stride_x` - Horizontal stride
+/// * `stride_y` - Vertical stride
+/// * `mac_clip` - Amount of bits clipped at the end of the convolution
+/// * `pp_clip` - Amount of bits clipped at the end of the post processing
+
 #[no_mangle]
 pub unsafe extern "C" fn dla_conv2d_bias(
     input_data: *const i8,
     kernel_data: *const i8,
-    // NOTE: bias is actually i16 in hardware, here we use 32 for TVM compatibility
+    // NOTE:
     bias: *const i32,
     output: *mut i8,
     input_channels: usize,
@@ -351,7 +378,7 @@ pub unsafe extern "C" fn dla_conv2d_bias(
 ///
 /// * `input_data` - Input data  
 /// * `kernel_data` - Kernel/weight data
-/// * `bias` - Buffer containing bias data
+/// * `bias` - Buffer containing bias data. NOTE: Bias is actually i16 in hardware, here we use 32 for TVM compatibility
 /// * `output` - Buffer for operation output
 /// * `input_channels` - Number of channels in the input
 /// * `input_height` - Height of the input
@@ -459,7 +486,7 @@ pub unsafe extern "C" fn dla_conv2d_bias_relu(
 ///
 /// * `input_data` - Input data  
 /// * `kernel_data` - Kernel/weight data
-/// * `bias` - Buffer containing bias data
+/// * `bias` - Buffer containing bias data. NOTE: Bias is actually i16 in hardware, here we use 32 for TVM compatibility
 /// * `output` - Buffer for operation output
 /// * `output_scale` - Multiplier to scale output of requantization
 /// * `output_zero` - Zero point of the requantization output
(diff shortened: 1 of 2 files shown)


examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -20,17 +20,24 @@ unsafe fn ffi_data_import(
...
-    let input_data: Vec<i8> = unsafe {
+    let mut input_data: Vec<i8> = unsafe {
         slice::from_raw_parts(input_data, input_channels * input_height * input_width).to_vec()
     };
 
+    // Input zero point shift
+    input_data = input_data
+        .into_iter()
+        .map(|x| scale_as_i16(x, input_zero))
+        .collect();
```
hegza: Maybe more clearly expressed as:

```rust
for x in input_data.iter_mut() {
    scale_as_i16(x, input_zero)
}
```

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -205,17 +214,18 @@ pub unsafe extern "C" fn dla_conv2d_relu(
 #[no_mangle]
 pub unsafe extern "C" fn dla_conv2d_bias(
     input_data: *const i8,
+    kernel_data: *const i8,
+    bias: *const i32, // NOTE: bias is actually i16 in hardware, here we use 32 for TVM compatability
```
hegza: Add the comment above the relevant line, like this:
```rust
// Comment on bias here
bias: *const i32
```

If you ever have to parse source code or abstract syntax trees, I believe you'll see why this is more intuitive.

hegza: Consider also adding a full `Arguments` block on the function:

```rust
/// Do stuff
///
/// # Arguments
///
/// * `bias` - explain stuff
fn foo() { /* ... */ }
```

PS: *"compatibility"

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -312,12 +323,14 @@ pub unsafe extern "C" fn dla_conv2d_bias_relu(
             kernel_order,
         )
     };
-    let bias: Vec<i16> = unsafe { slice::from_raw_parts(bias, bias_length).to_vec() };
+
+    let bias: Vec<i32> = unsafe { slice::from_raw_parts(bias as *const i32, bias_length).to_vec() };
+    let bias_i16: Vec<i16> = bias.into_iter().map(|x| clip_i32_to_i16(x)).collect();
```
hegza: Useless allocation on first line. Remove `to_vec` to manage bias as a slice (a view into the original buffer) instead.

This actually raises another question about allocations: do these buffers need to be allocated in the first place? I think we could actually run the entire thing with barely any allocations by changing almost all `Vec`'s to slices (including inside the driver proper). But that's an optimization topic and not too important right now.

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -333,7 +346,139 @@ pub unsafe extern "C" fn dla_conv2d_bias_relu(
...
+            kernel_data,
+            kernel_amount,
+            kernel_channels,
+            kernel_height,
+            kernel_width,
+            kernel_order,
+        )
     };
+
+    let bias: Vec<i32> = unsafe { slice::from_raw_parts(bias as *const i32, bias_length).to_vec() };
```
hegza: Looks like some useless allocation going on here as well.

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -333,7 +346,139 @@ pub unsafe extern "C" fn dla_conv2d_bias_relu(
...
+    };
+}
+fn clip_i32_to_i16(value: i32) -> i16 {
+    if value > i16::MAX as i32 {
+        return i16::MAX;
+    } else if value < i16::MIN as i32 {
+        return i16::MIN;
+    } else {
+        return value as i16;
+    }
```
hegza: Consider the standard library function `value.clamp(i16::MIN as i32, i16::MAX as i32) as i16` which implements the same behavior. It does include "an extra" assert for checking `MAX > MIN`, but I'm quite confident the compiler gets rid of that since MIN & MAX are known at compile-time.

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -333,7 +346,139 @@ pub unsafe extern "C" fn dla_conv2d_bias_relu(
...
+}
+
+fn scale_as_i16(input: i8, factor: i16) -> i8 {
+    let new_value = input as i16 + factor;
+    if new_value > 127 {
+        return 127;
+    } else if new_value < -128 {
+        return -128;
+    }
```
hegza: Prefer symbolic values over magic numbers: `i8::MAX = 127`, `i8::MIN = -128`.

hegza: `i16::clamp` works here too.

hegza (changes requested):
Some clarifications required. Some suggestions for performance.

Also, I think some of the changes relate to the clipping hardware bug. Document very clearly if there is software code that is accounting for hardware problems.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Fixed

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Makes sense. Arguments blocks added.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Removed useless allocation

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Removed useless allocation

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Function removed and replaced by direct calls to clamps

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Function removed and calls replaced by suitable clamp

vilukissa68:
Added better documentation for the clipping bug fix

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -9,28 +9,49 @@ use alloc::vec::Vec;
...
+/// * `input_height` - Height of the input
+/// * `input_width` - Width of the input
+/// * `input_order` - Layout of the input data
+/// * `input_zero` - Possible zero point correction value
+/// * `kernel_data` - Kernel/weight data
+/// * `kernel_amount` - Number of output channels
+/// * `kernel_channels` - Number of input channels
+/// * `kernel_height` - Kernel height
+/// * `kernel_width` - Kernel width
+/// * `kernel_order` - Kernel data layout
```
hegza: That's a little excessive :sweat_smile:  I meant to document the ones that are unobvious from context. I think documenting that `kernel_height` means "Kernel height" is just unnecessary inertia.

hegza: Same goes for other repetitions of this pattern.

hegza (approved):
Good but check the comment.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Removed unnecessary documentation

PR #73, opened 2024-09-25
Fix kernel data writing to DLA in TVM order

vilukissa68:


vilukissa68:
During TVM testing I found out that the permutation of 4DArrays doesn't work as expected. Instead of checking all the permutations, I just made sure that the standard TVM order HWKC can be written to the DLA. Validation tests use the DLA standard already so they are not affected by the permutation not working. 

Realistically most of the orders are not used, so we should choose which orders to support and limit DLA driver to those.

examples/hpc/dla-driver/src/tensor4.rs
```diff
@@ -324,8 +321,38 @@ impl<T: Clone> Tensor4<T> {
...
+                            + i_idx * self.kernels()
+                            + o_idx;
+                        hwoi_flat.push(data[hwio_index].clone());
+                    }
+                }
+            }
+        }
+
+        hwoi_flat
+    }
```
hegza, line 357: I'm pretty sure `ndarray::Array4` already has a method to do something like this (maybe [permuted_axes()](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.permuted_axes)). So it might be easier to do that operation on `data`, then call `iter().to_vec()`.

hegza (approved):
Seems good. ndarray might have a useful method for permuted_axes but you decide whether it's good or not.

examples/hpc/dla-driver/src/tensor4.rs
vilukissa68: Yes, I tried to implement this functionality with `permuted_axes()`, but I observed that it didn't work as I would have liked. It might be something with my implementation. For now, I will leave this as it is, and come back to it later if there is a need for other orders to work.

PR #80, opened 2024-10-25
Feat/dla clip handling

vilukissa68:
Changes results reading from DLA to match TVM requantization pattern. Essentially scaling responsibility is moved to TVM graph transformations.

### DIFF
--- a/examples/hpc/dla-driver-ffi/src/lib.rs
+++ b/examples/hpc/dla-driver-ffi/src/lib.rs
@@ -4,6 +4,7 @@
 #![no_std]
 #![no_main]
 
+#[macro_use]
 extern crate alloc;
 use alloc::vec::Vec;
 use core::ffi::{c_char, CStr};
@@ -20,7 +21,6 @@ unsafe fn ffi_data_import(
     input_height: usize,
     input_width: usize,
     input_order: *const c_char,
-    input_zero: i16,
     kernel_data: *const i8,
     kernel_amount: usize,
     kernel_channels: usize,
@@ -31,11 +31,6 @@ unsafe fn ffi_data_import(
     let mut input_data: Vec<i8> = unsafe {
         slice::from_raw_parts(input_data, input_channels * input_height * input_width).to_vec()
     };
-
-    // Input zero point shift
-    for x in input_data.iter_mut() {
-        *x = i16::clamp((*x) as i16 + input_zero, i8::MIN as i16, i8::MAX as i16) as i8
-    }
 
     let input_order_string = unsafe { CStr::from_ptr(input_order).to_str().unwrap_unchecked() };
     let input_tensor = unsafe {
@@ -112,7 +107,6 @@ pub unsafe extern "C" fn dla_conv2d(
             input_height,
             input_width,
             input_order,
-            0,
             kernel_data,
             kernel_amount,
             kernel_channels,
@@ -177,7 +171,6 @@ pub unsafe extern "C" fn dla_conv2d_relu(
             input_height,
             input_width,
             input_order,
-            0,
             kernel_data,
             kernel_amount,
             kernel_channels,
@@ -248,7 +241,6 @@ pub unsafe extern "C" fn dla_conv2d_bias(
             input_height,
             input_width,
             input_order,
-            0,
             kernel_data,
             kernel_amount,
             kernel_channels,
@@ -322,7 +314,6 @@ pub unsafe extern "C" fn dla_conv2d_bias_relu(
             input_height,
             input_width,
             input_order,
-            0,
             kernel_data,
             kernel_amount,
             kernel_channels,
@@ -380,10 +371,10 @@ pub unsafe extern "C" fn dla_tvm_qnn_conv2d(
     kernel_data: *const i8,
     bias: *const i32,
     output: *mut i8,
-    output_scale: *const f32,
-    output_zero: *const i32,
     input_scale: *const f32,
     input_zero: *const i32,
+    output_scale: *const f32,
+    output_zero: *const i32,
     input_channels: usize,
     input_height: usize,
     input_width: usize,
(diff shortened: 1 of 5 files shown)


examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -459,28 +436,20 @@ pub unsafe extern "C" fn dla_tvm_qnn_conv2d(
         None,
     );
 
+    let input_order_string = unsafe { CStr::from_ptr(input_order).to_str().unwrap_unchecked() };
```
hegza, line 440: I guess TVM is always *guaranteed* to give valid order strings. This would be a great way inject a buffer underrun/overflow through the FFI :) But we don't much care about those for this so I think this is fine and efficient.

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -459,28 +436,20 @@ pub unsafe extern "C" fn dla_tvm_qnn_conv2d(
...
-        u32::pow(2, pp_clip) as f32,
-        input_zero[0],
-        output_zero[0],
-        input_scale[0],
-        output_scale,
-    );
-
-    let input_order_string = unsafe { CStr::from_ptr(input_order).to_str().unwrap_unchecked() };
+    let mut res_i32: Vec<i32> = result.to_buffer_with_order(Order3::try_from(input_order_string).unwrap_unchecked())
+                                       .iter().map(|x: &i8| (*x as f32 * u32::pow(2, pp_clip) as f32) as i32).collect();
```
hegza, line 449: By the way, if this is related to the hardware bug, one thing you could do is to `#[cfg(feature = "asic")]`. That way we can still get accurate results from the VP while addressing the issues with the ASIC.

hegza: OTOH it can be good to be as close to hardware implementation as possible with the VP as well, bugs and all included. idk, food for thought :)

examples/hpc/dla-driver/src/tensor3.rs
```diff
@@ -260,9 +260,8 @@ pub fn rescale(
         };
 
         channel_slice.map_inplace(|x| {
-            let value = (((*x as f32 * pre_scale) - output_zero as f32) / input_scale) * scale
-                + input_zero as f32;
-            *x = value.clamp(i8::MIN as f32, i8::MAX as f32) as i8
+             let value = (input_scale / scale) * (*x as f32 * pre_scale - input_zero as f32)
+                 + output_zero as f32;
         });
```
hegza: Assignment missing from `map_inplace`. I think this line is a no-op as it is. Strange you're not getting any complaints from the compiler or clippy :)

vp/devel/python_peripherals/DLA.py
```diff
@@ -895,7 +907,7 @@ def get_input_data(self):
                     column_wise.append(data[idx])
 
         column_wise = reshape(column_wise, (channels, height, width))
-        print_matrix(column_wise[0], "flat input:", pformat="decimal")
+        #print_matrix(column_wise[0], "flat input:", pformat="decimal")
```
hegza: Only-slight-frown upon the commented out print. I do realize there's not really a good way to configure whether a print shall be on or not with this infra, apart from commenting out.

vp/devel/python_peripherals/DLA.py
```diff
@@ -1153,62 +1168,61 @@ def conv2d(self, input_img, kernels, padding, dilation=(1,1), stride=(1,1), padd
...
+                        center_x = center_x_0 + (w * stride[0])
+                        range_x = [center_x + k * dilation[0] for k in range(-w_kernel_max_offset, w_kernel_max_offset + 1)]
                     else:
-                        center_y = center_y_0 + (h * stride[1]) # Calculate from top left of center
-                        range_y = [center_y + k * dilation[1] for k in range(0, h_kernel_max_offset + 1)]
+                        center_x = center_x_0 + (w * stride[0])
+                        range_x = [center_x + k * dilation[0] for k in range(-w_kernel_max_offset, w_kernel_max_offset)]
 
-                    # Sum each channel with current kernel
+                        # Sum each channel with current kernel
```
hegza: This line adds inconsistent whitespace.

hegza (changes requested):
I didn't look *too closely* but I did find a possible bug in there. I'd say no need to re-submit for review. Just fix & merge if you agree..

examples/hpc/dla-driver/src/tensor3.rs
hegza: Yeah I guess `map_inplace` doesn't give much in the way of static analysis. I think almost an equivalent but safer alternative would be to use `into_iterator` to consume the original allocation. This will likely result in the same assembly code but also not allow such issues as possibly introduced by this diff.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Yes this string is hard-coded to Headsail TVM BYOC backend. One could use this for malicious purposes but this is fast and easy for us. When NVIDIA buys us, I will fix this.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: It's not really a hardware bug, but rather the way hardware is designed. DLA-VP implements this hardware behavior to ensure compatibility with RTL tests. More on this in my thesis :D

examples/hpc/dla-driver/src/tensor3.rs
vilukissa68: Returned this to its original state, as the assignment was accidentally removed. The new codegen only hits this part of the code in an edge so it didn't cause problems during my testing.

vp/devel/python_peripherals/DLA.py
vilukissa68: I don't think these kinds of changes should be introduced as a part of a new feature anyways. This one accidentally slipped past. Change removed.

vp/devel/python_peripherals/DLA.py
vilukissa68: Whitespace fixed.

PR #81, opened 2024-10-30
feat: add the break & start hack

hegza:
Is this still relevant / useful? Should we merge this.


vilukissa68 (commented):
As it looks like this only fixes some programs, I don't think this should be merged.

hegza:
Closing on verdict.

PR #82, opened 2024-10-30
Add newlib artifacts (@v0.0.1)

hegza:
Origin: <https://gitlab.tuni.fi/soc-hub/headsail/sw/newlib-artifacts/-/tags/v0.0.1>

This diff is quite heavy, since it includes the binary artifacts as discussed.

One thing I'd like to discuss is that it seems that the artifact repository contains not just static libs and headers but also object files as well. I believe these should not be required and should be omitted. Would you agree?


hegza:
We might also want to specify the version of upstream newlib used. I didn't find that documented on the newlib artifacts repository.

hegza:
It seems that we may have to include a COPYING.NEWLIB file as well, since we're doing a binary distribution. That would have to be supplied in the artifacts repository as well.

hegza:
~~This PR seems to increase the CI runtime by 10×. Does not seem viable at face value. Its probably because fresh cloning the repo takes a minute instead of taking a second.~~ No I just can't interpret the CI right.

andstepan:
> This diff is quite heavy, since it includes the binary artifacts as discussed.
>
>One thing I'd like to discuss is that it seems that the artifact repository contains not just static libs and headers but also >object files as well. I believe these should not be required and should be omitted. Would you agree?

Well I am not entirely sure if that is the case. I mean, we **do** have a problem with the size, since it looks like artifacts that we do not need get included (like, for example, rv32imac). I think there is a way to configure Newlib to not do this and to just build one configuration. I can look into it if you @hegza  think that it's going to be worth it.

hegza:
Discussed offline, most likely is worth it to make the artifacts smaller. One alternative would be to compress these, and upload them as some other type of artifact into GitHub, then use the CI automation to download them as needed.

**Actually, it is most likely a very good idea to just `tar -cz` these files. Any local user or workflow can uncompress them as needed, and CI can also uncompress them reasonably fast.**

hegza:
One thing to note here is that we likely want **both** VP and ASIC variants. Perhaps rename directory to match: `newlib-vp` and `newlib-asic`. They of course should contain the hard-float and soft-float variants inside the appropriate subdirectory.

PR #87, opened 2024-10-31
Add missing things to REPL

hegza:
* feat: add SysCtrl uDMA peripherals & ITC
* fix: set PLIC nr. of sources to 27. This seems more reasonable, as the highest interrupt index is 27 (Ethernet1).


hegza:
Notably, the emulated uDMA UART doesn't seem to give output to the analyzer so far. Might be a problem with the REPL file.

PR #93, opened 2024-11-29
Add support for grouped convolution

vilukissa68:


hegza:
Build failure for `check_bsp` seems unrelated. I'll fix that when I can, this pipeline should be rebased & rerun after that. The rest of the failures seem relevant.

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -71,8 +72,7 @@ unsafe fn ffi_data_import(
 /// Initializes DLA by setting up necessary heap allocator from headsail-bsp. This should be called only once in the program.
 #[no_mangle]
 pub unsafe extern "C" fn dla_init() {
-    // SAFETY: `init_heap` must be called once only
-    unsafe { init_heap() };
```
hegza: Is this reverting a prior upstream change? Seems strange. I think init_heap is what it's supposed to be atm.

vilukissa68:
The majority of changes related to automatic formatting changes after the Emacs Python linter started working out of nowhere. You can ignore the Python changes.

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -28,7 +29,7 @@ unsafe fn ffi_data_import(
     kernel_width: usize,
     kernel_order: *const c_char,
 ) -> (Tensor3<i8>, Tensor4<i8>) {
```
hegza, line 32: You do have to wonder if some of these could or should be combined into structs even for the FFI, e.g.,

```c
typedef struct data_t {
  int8_t *data,
  size_t channels,
  size_t height,
  size_t width,
  char *order,
} data_t;`
```

```c
data_t input { .data = ... }
...
ffi_data_import(input, kernel)
```

Might not make sense to commit to refactoring at this point in the project though unless we have some reason to continue the work next year :)

examples/hpc/dla-driver-ffi/src/lib.rs
```diff
@@ -433,26 +440,119 @@ pub unsafe extern "C" fn dla_tvm_qnn_conv2d_bias(
             y: stride_y,
         }),
         Some(mac_clip),
-        Some(pp_clip),
+        Some(optimized_pp),
         None,
     );
 
-    let input_order_string = unsafe { CStr::from_ptr(input_order).to_str().unwrap_unchecked() };
+    let _input_order_string = unsafe { CStr::from_ptr(input_order).to_str().unwrap_unchecked() };
```
hegza: Unused? Just remove it.

examples/hpc/dla-driver/examples/depthwise.rs
```diff
@@ -0,0 +1,99 @@
...
+use panic_halt as _;
+
+use dla_driver::tensor3::{Order3, Tensor3};
+use dla_driver::tensor4::{Order4, Tensor4};
+
+use alloc::vec::Vec;
+
+fn conv_test() {
+    sprintln!("conv_test: enter");
+    let din: Vec<i8> = vec![
```
hegza, line 19: Tag this entry with `#[rustfmt::skip]` to make sure it doesn't get autoformatted.

examples/hpc/dla-driver/src/layers.rs
```diff
@@ -179,6 +179,50 @@ pub fn conv2d_bias_relu<T: DlaOutput + Clone>(
     )
 }
 
+pub fn grouped_conv2d<T: DlaOutput + Clone>(
```
hegza, line 265: Public API function: consider documenting what the function is trying to achieve. Possibly even `# Arguments` as well.

examples/hpc/dla-driver/src/tensor3.rs
```diff
@@ -183,6 +183,56 @@ impl<T: Clone> Tensor3<T> {
         self.order
     }
 
+    /// Concatenates a Tensor along the least significant axis (axis=2) by interleaving the tensors
+    pub fn concat_interleaved(tensors: Vec<Tensor3<T>>) -> Tensor3<T> {
```
hegza: This function does not seem to require ownership of the input buffer `tensors`. Consider only borrowing a slice for the duration of the function call: `concat_interleaved(tensors: &[Tensor3<T>])`. You'll need to add the borrow symbol `&` on the callsite as well.

hegza: Similar thing might apply to other functions around here, I didn't review those.

hegza (approved):
Nothing too crazy here. It's quite a lot of code for a single review though, so please try to keep in mind you can and should push formatting changes, whitespace, typos, etc. separately first (perhaps directly to main), then put a PR for review with the interesting changes.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Changed to use call directly with bsp

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Yes, if I or someone else wants to revisit this for optimizations at some point, this should be done. However, it's hard to justify now.

examples/hpc/dla-driver-ffi/src/lib.rs
vilukissa68: Removed

examples/hpc/dla-driver/examples/depthwise.rs
vilukissa68: Tagged

examples/hpc/dla-driver/src/layers.rs
vilukissa68: API documentation added

examples/hpc/dla-driver/src/tensor3.rs
vilukissa68: I don't understand the actual difference between a vec and a slice, but this works alright with this change. You probably know better :D

hegza: @vilukissa68  highlighting for resurrect.

You could use borrowed Vec as well but borrowed slice (&[T]) is just more general than Vec. It reads: "a view to any contiguous memory" without specifying if it should be allocated as a Vec specifically, ... or array, or Hashmap. This allows the function to be called with any of the above types.

PR #94, opened 2024-12-12
Add strict align for HPC builds

hegza:


hegza:
I don't know if this actually works. We should find some way to confirm that.

hegza:
Closing as per #95

PR #99, opened 2025-03-11
DLA: fix output width according to whether code runs in CI or not

andstepan:
Changes output data width according to whether the VP runs in CI or not.

### DIFF
--- a/vp/devel/python_peripherals/DLA.py
+++ b/vp/devel/python_peripherals/DLA.py
@@ -1183,7 +1183,7 @@ class Dla:
                 res = execute_for_all_elements(self.mac.relu_native, res)
 
         # Prevent overflowing i16 range
-        if output_bit_width == 32:
+        if output_bit_width == 32 and os.environ.get("CI"):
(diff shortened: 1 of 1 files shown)


vp/devel/python_peripherals/DLA.py
```diff
@@ -1183,7 +1183,7 @@ def next(element):
                 res = execute_for_all_elements(self.mac.relu_native, res)
 
         # Prevent overflowing i16 range
-        if output_bit_width == 32:
+        if output_bit_width == 32 and os.environ.get("CI"):
```
vilukissa68, line 1187:
```suggestion
# NOTE: (20250312 vaino-waltteri.granat@tuni.fi) Some CI tests expect 32-bit output, which is not supported on ASIC but is supported by VP. Make this option only available in CI.
        if output_bit_width == 32 and os.environ.get("CI"):
```
Explanation for the behaviour.

vilukissa68 (commented):
I added a note suggestion for the changed line.

